PR #1, opened 2019-08-07
Addresses HT-1196 update static IP

moseshll:
Add edit page and a whole lot of plumbing so that we have actual tests.

### DIFF
--- a/app/controllers/ht_users_controller.rb
+++ b/app/controllers/ht_users_controller.rb
@@ -1,16 +1,38 @@
 # frozen_string_literal: true
 
 class HTUsersController < ApplicationController
+  before_action :fetch_user, only: %i[show edit update]
+
   def index
     if params[:email]
       @users = HTUser.where('email LIKE ?', "%#{params[:email]}%").order(:userid)
+      flash.now[:alert] = "No results for '#{params[:email]}'" unless @users.count.positive?
     else
       @users = HTUser.all.order(:userid)
     end
   end
 
+  def update
+    if @user.update(user_params)
+      flash[:notice] = 'User updated'
+      redirect_to @user
+    else
+      flash.now[:alert] = @user.errors.full_messages.to_sentence
+      @user = HTUser.find(Base64.decode64(params[:id]))
+      render 'edit'
+    end
+  end
+
+  private
+
   # ht_users.userid entries can include periods and @ signs, which messes up routing.
-  def show
+  # So we wrap it in Base64 wherever we need to pass it around.
+  # A better solution would be to modify the ht_users schema to use integral primary key.
+  def fetch_user
     @user = HTUser.find(Base64.decode64(params[:id]))
+  end
+
+  def user_params
+    params.require(:ht_user).permit(:userid, :iprestrict)
   end
 end
--- a/app/controllers/ht_users_controller.rb
+++ b/app/controllers/ht_users_controller.rb
@@ -6,7 +6,7 @@ class HTUsersController < ApplicationController
   def index
     if params[:email]
       @users = HTUser.where('email LIKE ?', "%#{params[:email]}%").order(:userid)
-      flash.now[:alert] = "No results for '#{params[:email]}'" unless @users.count.positive?
+      flash.now[:alert] = "No results for '#{params[:email]}'" if @users.empty?
     else
       @users = HTUser.all.order(:userid)
     end
--- a/app/models/ht_user.rb
+++ b/app/models/ht_user.rb
@@ -27,6 +27,7 @@ class HTUser < ApplicationRecord
   end
 
   def iprestrict=(val)
+    val = val.strip
     escaped = '^' + val.gsub('.', '\.') + '$'
     write_attribute(:iprestrict, escaped)
   end
--- a/app/models/ht_user.rb
+++ b/app/models/ht_user.rb
@@ -1,9 +1,8 @@
 # frozen_string_literal: true
 
 class HTUser < ApplicationRecord
-  # Brain-dead IPv4 validation
   validates :iprestrict, presence: true,
-                         format: { with: /\A\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}\z/,
+                         format: { with: /\A(?:(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\z/,
                                    message: 'requires a valid IPv4 address' }
 
   HUMANIZED_ATTRIBUTES = {
--- a/app/models/ht_user.rb
+++ b/app/models/ht_user.rb
@@ -1,5 +1,34 @@
 # frozen_string_literal: true
 
 class HTUser < ApplicationRecord
-  # self.table_name = "ht_users"
+  # Brain-dead IPv4 validation
+  validates :iprestrict, presence: true,
+                         format: { with: /\A\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}\z/,
+                                   message: 'requires a valid IPv4 address' }
+
+  HUMANIZED_ATTRIBUTES = {
+    iprestrict: 'IP Restriction'
+  }.freeze
+
+  def self.human_attribute_name(attr, options = {})
+    HUMANIZED_ATTRIBUTES[attr.to_sym] || super
+  end
+
+  def to_param
+    Base64.encode64(userid)
+  end
+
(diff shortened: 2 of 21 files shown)


test/models/ht_user_test.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+require 'byebug'
+
+class HTUserTest < ActiveSupport::TestCase
```
aelkiss: I'd like to see some testing of the escaping/unescaping here.

moseshll: Test added

app/models/ht_user.rb
```diff
@@ -1,5 +1,34 @@
 # frozen_string_literal: true
 
 class HTUser < ApplicationRecord
-  # self.table_name = "ht_users"
+  # Brain-dead IPv4 validation
+  validates :iprestrict, presence: true,
+                         format: { with: /\A\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}\z/,
```
billdueber: Check out https://www.oreilly.com/library/view/regular-expressions-cookbook/9780596802837/ch07s16.html for a battle-tested ip regex

moseshll: Battle-tested regex swapped in, except ^ and $ replaced by \A and \z to keep Ruby happy.

app/views/shared/_flash_message.html.erb
```diff
@@ -0,0 +1,13 @@
+<% unless flash.blank? %>
+  <% flash.each do |key, value| %>
+    <% alert_class = case key
```
billdueber: <priority very low> This seems like another place where we could use a translation file and use something like t("alert.css_class.#{key}", default: "alert-"#{key}"). In any case, it doesn't feel to me like it belongs here.

db/schema.rb
```diff
@@ -0,0 +1,31 @@
...
+# database schema. If you need to create the application database on another
+# system, you should be using db:schema:load, not running all the migrations
+# from scratch. The latter is a flawed and unsustainable approach (the more migrations
+# you'll amass, the slower it'll run and the greater likelihood for issues).
+#
+# It's strongly recommended that you check this file into your version control system.
+
+ActiveRecord::Schema.define(version: 0) do
+
+  create_table "ht_users", primary_key: "userid", id: :string, limit: 256, default: "", options: "ENGINE=MyISAM DEFAULT CHARSET=latin1", force: :cascade do |t|
```
billdueber: @aelkiss Are we not defaulting to innodb these days? Or not so much?

db/schema.rb
```diff
@@ -0,0 +1,31 @@
...
+    t.string "displayname", limit: 128
+    t.string "email", limit: 128
+    t.string "activitycontact", limit: 128
+    t.string "approver", limit: 128
```
billdueber: If the userid can be 256, don't these need to be as well?

app/models/ht_user.rb
billdueber: I'd use `\Z` (ignores trailing newline) and maybe add some `\s*` on the front and back so we don't break on leading/trailing spaces.

app/views/ht_users/index.html.erb
```diff
@@ -1,32 +1,34 @@
...
+    <th>IP Restriction</th>
+    <th>MFA?</th>
+
+    <% @users.each do |u| %>
+      <tr>
+        <td><%= link_to u.userid, ht_user_path(u) %></td>
+        <td><%= u[:displayname] %></td>
+        <td><%= u[:email] %></td>
+        <td><%= u[:role] %></td>
+        <td><%= u[:expires]&.strftime("%Y-%m-%d") %></td>
```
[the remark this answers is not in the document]
moseshll: Artifact of development with placeholder DB contents, IIRC. The schema says NOT NULL so it's not likely to be an issue one way or the other.

app/models/ht_user.rb
moseshll: Comments distilled into an edit of HT-1494.
Added a strip() to deal with whitespace for the current PR.

app/models/ht_user.rb
```diff
@@ -1,5 +1,34 @@
...
+    Base64.encode64(userid)
+  end
+
+  # iprestrict is in the database as an escaped IPv4 regex e.g., ^127\.0\.0\.1$
+  # For the UI we strip out and restore the ^/./$ characters
+  def iprestrict
+    escaped = self[:iprestrict]
+    return escaped if escaped.nil?
+
+    escaped.gsub(/^\^/, '').gsub(/\$$/, '').gsub(/\\\./, '.')
```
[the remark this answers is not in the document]
moseshll: I think a full-blown regex was used so A&E could accommodate IP ranges. I'm thinking in particular of one I saw for the Copyright Office a while back. Aaron could probably dispel any myths I may be harboring.

app/controllers/ht_users_controller.rb
```diff
@@ -1,16 +1,38 @@
...
   def index
     if params[:email]
       @users = HTUser.where('email LIKE ?', "%#{params[:email]}%").order(:userid)
+      flash.now[:alert] = "No results for '#{params[:email]}'" unless @users.count.positive?
     else
       @users = HTUser.all.order(:userid)
     end
   end
 
+  def update
```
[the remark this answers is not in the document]
moseshll: Leaving that up to Jira.

app/controllers/ht_users_controller.rb
```diff
@@ -1,16 +1,38 @@
 # frozen_string_literal: true
 
 class HTUsersController < ApplicationController
+  before_action :fetch_user, only: %i[show edit update]
+
   def index
     if params[:email]
       @users = HTUser.where('email LIKE ?', "%#{params[:email]}%").order(:userid)
+      flash.now[:alert] = "No results for '#{params[:email]}'" unless @users.count.positive?
```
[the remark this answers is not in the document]
moseshll: Modified

app/models/ht_user.rb
```diff
@@ -1,5 +1,34 @@
 # frozen_string_literal: true
 
 class HTUser < ApplicationRecord
-  # self.table_name = "ht_users"
+  # Brain-dead IPv4 validation
+  validates :iprestrict, presence: true,
+                         format: { with: /\A\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}\z/,
+                                   message: 'requires a valid IPv4 address' }
+
+  HUMANIZED_ATTRIBUTES = {
```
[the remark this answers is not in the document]
moseshll: HT-1495 added.

app/views/shared/_flash_message.html.erb
moseshll: twitter-bootstrap-rails gem might take care of this for us.

aelkiss:
Regarding the regexes for IP ranges: The escaping of "." is required because "." matches any character by default in a regex. I agree the reason for the regexes is so that we can handle ranges or (potentially) multiple individual IP addresses for a person. That's definitely not the best way to deal with it, but changing the schema for the ht_users table is definitely out of scope here though as it would affect other components. I suspect we will eventually want to overhaul how we do user management entirely and at that time we can think about storing the IP ranges in a more robust way.

db/schema.rb
aelkiss: I believe that we are. I think it's worth noting that db:setup would _never_ be run in production with this app - the table is created externally and this app just manages it.

db/schema.rb
aelkiss: userid could be a shib persistent ID which is probably why it's longer. This schema is just copied from production, and changes to it are out of scope for this project.

app/views/ht_users/index.html.erb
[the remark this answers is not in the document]
aelkiss: See comment below about regexes: I believe that we are at least currently, and changes to that are out of scope right now (but worth keeping in mind for if/when we redesign user management entirely)

app/models/ht_user.rb
```diff
@@ -1,5 +1,34 @@
...
+    iprestrict: 'IP Restriction'
+  }.freeze
+
+  def self.human_attribute_name(attr, options = {})
+    HUMANIZED_ATTRIBUTES[attr.to_sym] || super
+  end
+
+  def to_param
+    Base64.encode64(userid)
```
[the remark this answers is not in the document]
moseshll: For some reason I can't make this work with /[^\/]+/ but it does work with /.+/ which seems like a nastier hack. Not going to try to modify this PR with it. It's HT-1497.

PR #10, opened 2019-08-23
HT-1506: Configure relative url root

aelkiss:
This sets up relative url root handling that works in dev.

It does not work in test with a relative url root of anything other than '/' (among other things it complains there are no matching routes) for reasons I don't care to debug at the moment.

This also sets up default dev & test database config so that it can be used locally out of the box without requiring a mysql setup.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ before_install:
   - gem install bundler
 
 before_script:
+  - bundle exec rake keycard:migrate RAILS_ENV=test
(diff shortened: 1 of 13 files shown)


aelkiss:
> I advocate that we avoid using ` Otis.config.relative_url_root` directly wherever possible. In this case, that would mean replacing `Otis.config.relative_url_root` with `root_path` wherever it appears in the session controller.

Done.

PR #12, opened 2019-08-29
Addresses HT-1497 use a routing constraint instead of Base64 for userid.

moseshll:


aelkiss:
With `AllowEncodedSlashes nodecode` I get a 404 from the Rails app instead of from Apache, probably because the Rails app isn't decoding them...

aelkiss:
Same behavior with `AllowEncodedSlashes on`. I'll leave that on for now; feel free to take a look from the app side and see what you can get.

moseshll:
Sounds to me like we are stuck with Base64, or with using the `email` field as though it were the primary key.

aelkiss:
It's probably fixable. I would be OK just using the email field as a primary key. It certainly needs to be unique; effectively that's what we do now when referencing users outside the application.

moseshll:
Trying again, with email instead of userid.

PR #13, opened 2019-09-12
HT-1529 integrate shib

moseshll:
Not sure about where the Shib settings should live, or how they need to be tweaked further.

The second commit has also been deployed on Staging.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -8,7 +8,7 @@ class ApplicationController < ActionController::Base
 
   helper_method :logged_in?, :current_user
 
-  before_action :safe_validate_session
+  before_action :validate_session
   before_action :authenticate!
   before_action :authorize!
   protect_from_forgery with: :exception, unless: -> { authentication.csrf_safe? }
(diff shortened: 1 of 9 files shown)


aelkiss:
Also, just for the future, note that the github / jira integration will only work if you include the full issue key e.g. HT-1529 in the commit messages / PR title / branch name.

botimer:
@moseshll Hmm, yeah, it's odd to hear that the session is going away. I'd like to see that in action to understand what's happening and determine whether we need a keycard change.

aelkiss:
I thought maybe this was just a side effect of hitting the FakeShibController, but it looks like it's happening even in with staging. See the branch https://github.com/hathitrust/otis/tree/regular-validate-session (currently deployed to staging).

Based on what I'm seeing in dev, I suspect a timestamp isn't getting added to the session when the session is reset, but I will take a closer look and see.

aelkiss:
hmm... Now I'm unsure. It looks like Keycard's `setup_session` method should be preserving the `return_to` after coming back from login. I'll see if I can replicate in dev (with the additional caveat that FakeShibController also needs to preserve the `return_to`)

aelkiss:
Hmmm so I think `SessionController#new` tries to validate the session (thereby whacking the `return_to`) before issuing the redirect to Shibboleth.

botimer:
Yup... The session controllers are always sort of ring zero. `skip_before_action validate_session` is the right path.

PR #14, opened 2019-09-27
Addresses HT-1455 display institution name.

moseshll:
Part of the ticket relates to grouping by institution for batch operations. Since sorting by columns is not implemented yet, I'm taking that as not part of the MVP for the current effort.

### DIFF
--- a/app/models/ht_institution.rb
+++ b/app/models/ht_institution.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+# We're not really interested in editing or viewing this table,
+# but we want to allow HTUser access to a human-readable version of
+# the institution.
+class HTInstitution < ApplicationRecord
+  self.primary_key = 'sdrinst'
+  has_many :ht_users, foreign_key: :identity_provider, primary_key: :entityID
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -34,8 +34,8 @@ ActiveRecord::Schema.define(version: 0) do # rubocop:disable Metrics/BlockLength
   end
 
   create_table :ht_institutions, id: false do |t|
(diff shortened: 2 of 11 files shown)


app/models/ht_institution.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+# We're not really interested in editing or viewing this table,
+# but we want to allow HTUser access to a human-readable version of
+# the institution.
+class HTInstitution < ApplicationRecord
+  self.primary_key = 'sdrinst'
```
aelkiss: `sdrinst` is deprecated; use `inst_id` instead.

db/schema.rb
```diff
@@ -31,4 +33,24 @@
     t.string :identity_provider
   end
 
+  create_table :ht_institutions, id: false do |t|
+    t.string :sdrinst, primary_key: true
+    t.string :inst_id
```
aelkiss: As above, `sdrinst` is deprecated

aelkiss:
I'll also deploy and check that this works as expected in dev.

app/models/ht_institution.rb
moseshll: Should be addressed in second commit. Tests pass locally.

db/schema.rb
moseshll: Ditto above.

moseshll:
I'm not sure this will work if deployed to dev with the Shib branch unmerged.

aelkiss:
Works so long as you use the WAYFless link (that I had posted a while back in Slack) to force login. The column widths could be better but this will work for now. I'll go ahead and merge it.

PR #15, opened 2019-09-30
Addresses HT-1198 update user as expired.

moseshll:
Branch name is a lie, it's a button that pulls in the current timestamp as the user's expiration date.

### DIFF
--- a/test/models/ht_user_test.rb
+++ b/test/models/ht_user_test.rb
@@ -3,13 +3,13 @@
 require 'test_helper'
 
 class HTUserTest < ActiveSupport::TestCase
-  test 'iprestrict validation passes' do
-    user = HTUser.new(userid: 'user', iprestrict: '127.0.0.1')
+  test 'validation passes' do
+    user = create_test_ht_user('user')
     assert user.valid?
   end
 
   test 'iprestrict validation fails' do
-    user = HTUser.new(userid: 'user', iprestrict: '127.0.0.1.1')
+    user = create_test_ht_user('user', iprestrict: '127.0.0.1.1')
     assert_not user.valid?
   end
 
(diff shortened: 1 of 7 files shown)


test/models/ht_user_test.rb
```diff
@@ -3,13 +3,16 @@
...
+  valid_iprestrict = '127.0.0.1'
+
+  test 'validation passes' do
+    user = HTUser.new(userid: 'user', iprestrict: valid_iprestrict, expires: valid_expires)
     assert user.valid?
   end
 
   test 'iprestrict validation fails' do
-    user = HTUser.new(userid: 'user', iprestrict: '127.0.0.1.1')
+    user = HTUser.new(userid: 'user', iprestrict: '127.0.0.1.1', expires: valid_expires)
```
aelkiss: Having to add parameters to the constructor each time you add a required field seems like a pain. I would implement a factory method in your helpers (either yourself or using something like FactoryBot or Fabricator) that has `valid_expires` and `valid_iprestrict` as default parameters but lets you override them. Let me know if you need a concrete example of that.

app/views/ht_users/show.html.erb
```diff
@@ -14,7 +14,7 @@
       <dt>User Type:</dt> <dd><%= @user.usertype %></dd>
       <dt>Role:</dt> <dd><%= @user.role %></dd>
       <dt>Access:</dt> <dd><%= @user.access %></dd>
-      <dt>Expires:</dt> <dd><%= @user.expires %></dd>
+      <dt>Expires:</dt> <dd><%= @user.expires.to_s(:db) %></dd>
```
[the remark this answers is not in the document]
moseshll: For now I'll be happy with treating the expiration date sort of like we do the IP regex, with an HTUser .expires() method to ditch the 'UTC' suffix (which is what the .to_s(:db) does).

test/models/ht_user_test.rb
moseshll: Rebased against your comments and force-pushed.

PR #17, opened 2019-11-07
HT-1457: Sort user display by institution.

aelkiss:
- Was using "includes" to do an outer join, but that resulted in not
getting an institution name when the sort condition was added, for
reasons that are not clear to me (seems to be some kind of ActiveRecord
issue since the SQL query works fine & returns the institution name.)
The outer join shouldn't be necessary since all users will have an
identity provider that maps to an institution (or they'd be unable to
log in anyway.) A user that got added with a non-existent identity
provider means there's a bug somewhere. It might not hurt to be able to
audit for those but I think it's OK if they don't show up in the regular
display.

- This is reaching the limit of what can be done easily with simple factory
methods & the default rails fixtures; should consider using FactoryBot
or Fabricator.


aelkiss:
This is technically implemented, but probably needs a test in `ht_users_controller_test` and some cleanup for the factories as mentioned above.

aelkiss:
Now includes tests for sorting by institution as well as for HT-1459 (active vs. expired users), and removes the fixtures and helper function in favor of FactoryBot.

PR #18, opened 2019-11-12
HT-1875: Use FactoryBot instead of fixtures

aelkiss:
- Uses factories in ht_user_test and ht_users_controller_test
- Fixes direction of association for HTUser and HTInstitution
- Removes fixtures


aelkiss:
Going to merge this into #17

PR #23, opened 2019-12-02
HT-1458 add about to expire column

billdueber:
HT-1458 Add "about to expire" column …

Adds a column that shows a red flag if someone is expiring soon,
along with how many days until that happens.

TODO: Make the appropriate change to the production settings file
to set `Otis.config.expires_soon_in_days` (default: 30)

Notable:
  * Days until expiration is rounded to the nearest day (the underlying data is a full timestamp)
  * Users are considered to be "#expiring soon?" if they expire
    within `Otis.config.expires_soon_in_days`
    (set in config/settings/[env].yml)

### DIFF
--- a/app/models/ht_user.rb
+++ b/app/models/ht_user.rb
@@ -50,7 +50,30 @@ class HTUser < ApplicationRecord
     short ? self[:expires]&.strftime('%Y-%m-%d') : self[:expires]&.to_s(:db)
   end
 
+  # How many days until expiration?
+  # @return [Number] days until expiration
+  def days_until_expiration
+    seconds_until_expiration = self[:expires] - Time.zone.now
+    days_from_seconds(seconds_until_expiration)
+  end
+
+
+  # Is this person expiring "soon" (based on the config)?
+  # @return [Boolean]
+  def expiring_soon?
+    days_until_expiration.between? 0, (Otis.config&.expires_soon_in_days || 30)
+  end
+
   def institution
     ht_institution&.name
+  end
+
+  private
+  # Convert seconds (what we get when subtracting one date from another)
+  # to days
+  # @param [Number] secs How many seconds
+  # @return [Fixnum] How many days that represents, rounded
+  def days_from_seconds(secs)
+    (secs / (24*60*60)).to_i
   end
 end
--- a/app/models/ht_user.rb
+++ b/app/models/ht_user.rb
@@ -5,8 +5,8 @@ class HTUser < ApplicationRecord
   belongs_to :ht_institution, foreign_key: :identity_provider, primary_key: :entityID
 
   validates :iprestrict, presence: true,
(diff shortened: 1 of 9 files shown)


billdueber:
Screenshot:

image

config/settings/development.yml
```diff
@@ -1,6 +1,9 @@
 allow_impersonation: true
 relative_url_root: '/useradmin'
 
+# How many days away does expiration need to be before we start to worry?
+expires_soon_in_days: 30
```
aelkiss: Any reason not to make this the default in `config/settings.yml` and omit it here and in `config/settings/test.yml`?

test/models/ht_user_test.rb
```diff
@@ -62,3 +62,25 @@ def setup
...
+    assert_in_delta(-10, @expired_user.days_until_expiration, 1)
+    assert_in_delta(100, @safe_user.days_until_expiration, 1)
+  end
+
+  test '#expiring_soon is correct for those in and outside the range' do
+    assert(@expiring_user.expiring_soon?)
+    assert_not(@expired_user.expiring_soon?)
+    assert_not(@safe_user.expiring_soon?)
+  end
+end
```
aelkiss: Perhaps consider also adding a test (in ht_users_controller_test) that the expiring soon column is displayed?

app/models/ht_user.rb
```diff
@@ -50,7 +50,30 @@ def expires(short: false)
     short ? self[:expires]&.strftime('%Y-%m-%d') : self[:expires]&.to_s(:db)
   end
 
+  # How many days until expiration?
+  # @return [Number] days until expiration
+  def days_until_expiration
+    seconds_until_expiration = self[:expires] - Time.zone.now
```
aelkiss: Two issues:

1) In development at least, I get a `String` for self[:expires], not a `Time` (despite the column being a datetime in sqlite3...); haven't checked with mysql in production.
2) If one does get a Time, is there a reason not to use `self[:expires].to_date - Date.today`, and omit the `days_from_seconds` helper method? Alternatively, `Date.parse(self[:expires]) - Date.today`, if you actually get a string even in production...

billdueber: 1. It's actually a rails ActiveSupport::TimeWithZone object. They've change `#inspect`, `#pretty_print`, `#to_s`, etc. to always show the string representation, but it's a real timzone object underneath.

2. For two dates, d1 - d2 gives a Rational, but I can obviously cast that to an integer. I'll do that, and the other suggested changes on this PR.

aelkiss: Ah, thanks... I was looking with `pry`, but didn't look close enough...

aelkiss:
Tested at https://moseshll.babel.hathitrust.org/useradmin-staging/, looks good, merging..

PR #28, opened 2019-12-19
Build an actual object for the expiration_date

billdueber:
(This touches a *lot* of files, but only `app/models/expiration_date.rb` 
and the associated test file should need more than a cursory review)

The expiration date is being manipulated enough that
it's worth breaking out into its own class that
can compute things (e.g., whether it's expired,
how long until it expires, etc.)

An expiration date is both the actual 
date and the expires type (e.g, "expiresannually"). Expiration dates are
*compared* only by date.

Expiration dates are always dates only; the
time is thrown away. A quick question to Melissa
verified that this is ok.
https://hathitrust.slack.com/archives/DKV93G37T/p1576770049002400

The HTUser object now creates a new expiration date
object on creation and whenever `expires` is set and
forwards appropriate methods to it. The
expiration date object itself is read-only.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -35,3 +35,6 @@ Style/ConditionalAssignment:
 
 Style/Documentation:
   Enabled: false
+
+Layout/AlignHash:
+  Enabled: false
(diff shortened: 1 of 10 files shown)


aelkiss:
FWIW a similar kind of change might be useful for the IP address restriction. Let's rebase this and see about getting it merged soon (maybe worth waiting until after #38  is merged at least)

PR #30, opened 2020-01-24
Addresses HT-2007 iprestrict should be allowed to be NULL if MFA = true

moseshll:
Main change is the iprestrict validation being turned off if mfa is true. Changes to DB seeds and test factory to create mfa = true records.

Minor change to iprestrict= method to allow setting to nil, not sure if that is something that would happen in a production environment but I needed it (apparently) for the ht_user_mfa factory.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -23,6 +23,10 @@ Metrics/AbcSize:
 Metrics/LineLength:
   Enabled: false
 
+Metrics/ClassLength:
+  Exclude:
+    - 'test/**/*'
+
 Style/ClassAndModuleChildren:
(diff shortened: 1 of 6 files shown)


aelkiss:
lgtm; I'll test in dev Monday before merging, though. Thanks for the quick work.

aelkiss:
I didn't get to this today, but if it needs any changes I can take care of it and have Bill review while you're out on vacation.

aelkiss:
If a user with iprestrict = NULL is edited, then iprestrict gets set to the empty string, which does not match the current convention for users without ip restriction. I'll see what I can do.

PR #32, opened 2020-02-10
Addresses HT-1201 Should be controllable whether a user is on MFA

moseshll:
tl;dr add MFA checkbox to edit page.

More JQuery than I would like to have written for such a simple feature. Trying to avoid the situation where there is an iprestrict and mfa in the same entry, since they are mutually exclusive. Hence the disabling and setting iprestrict value to empty string and back when the user is noodling with the checkbox.

The form submit patch is so a disabled iprestrict field makes it into the request params so it can be set to nil.


aelkiss:
Once #38  is merged I'll rebase and re-review this.

aelkiss:
I can't currently think of cases where we would want to set mfa=1 but keep the iprestrict. Merging.

PR #36, opened 2020-04-16
Update README.md with local user config

moseshll:
Add needed setup step with config/settings.local.yml so the login step works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ bundle exec rake test
 ### 4. Trying it out
 
 ```
+echo "users: ['somebody@default.invalid']" > config/settings.local.yml
 bundle exec rails s
 ```
 


moseshll:
Actually, config/settings/development.yml has somebody@default.invalid in the users list, but it is not getting picked up. Additionally, I am having tests fail because the nobody@example.com user in config/settings/test.yml is apparently not being recognized in the sign_in! test helper. Guessing these point to the same problem. I created config/settings/test.local.yml with the test user and that worked. So there is some problem running out of the box for me, and I don't think this PR really is the solution. Not sure why Travis is OK with the current setup by my environment is not.

aelkiss:
I'm a little confused. The tests work out of the box for me. If you add a user to `settings.local.yml`, that will override whatever is in development/test.yml. The load order is documented here: https://github.com/mlibrary/ettin

moseshll:
I deleted all my local settings and we're back to working out of the box. No idea what happened there. A stray edit to a config file? Anyway, closing PR as bogative.

PR #38, opened 2020-04-17
HT-1494 OTIS must be able to edit and display multiple IP addresses

moseshll:


aelkiss:
Thought I had commented on this but I guess the comment disappeared. I'm thinking we're getting to the point where it would make sense to extract the `IPRestriction` as a separate class (similar to what Bill did in #28). I'm OK with going ahead and merging this but when we implement the ability to remove the IP address restriction (which we definitely need) -- i.e. setting it to `^.*$` -- we should for sure extract `IPRestriction` to a separate class that can convert between the input, storage (i.e. regex) and display versions.

PR #40, opened 2020-05-07
Bootstrap 3 adjustments and nav menu in header.

moseshll:
Pulled this out of the approval request branch since it will be useful to wire some other branches into this (e.g., UI for ht_institutions I am also working on).


billdueber:
@moseshll This looks good to me. You might also want to check out the relatively-ancient branch `better-look-merge` where I experimented with a full-on switch to boostrap-4.

moseshll:
@billdueber Thanks for the pointer, I will refer to that branch in future. I would like to just go with BS 4 but don't want to take the time right now.

PR #46, opened 2020-05-18
Settle on Time.zone.now instead of Time.now across the board.

moseshll:
This is an attempt to standardize on Time.zone.now etc since Time.now contributed to a test (approval_controller_test.rb) that failed at certain times of day due to a next-day GMT time getting stored in the database.


moseshll:
I'll resolve these conflicts when back from VAC ≥ 6/1

PR #50, opened 2020-05-21
HT-1585: fixes for one-time links

aelkiss:
See notes on individual commits.

I've verified on useradmin-testing that I can:

- generate an approval request
- send the email
- receive the email
- use the link in the email when logged in as a non-privileged user (after changing the url to development)


aelkiss:
Appeased rubocop and added a fix for HT-2381 as well.

PR #51, opened 2020-05-21
HT-2115 and HT-1940 renewal for multiple users.

moseshll:
As demonstrated to HT staff 2020-05-21.
Issue with excessively sticky flash message fixed.
Issue with existing approval request not being updated from approver -> HT staff not addressed because we don't know the extent of out-of-band communication with the approver, so for now let's leave the approver's identity intact on the request, even if (in the unlikely event) the request e-mail was never sent.

### DIFF
--- a/app/controllers/ht_approval_requests_controller.rb
+++ b/app/controllers/ht_approval_requests_controller.rb
@@ -3,18 +3,26 @@
 class HTApprovalRequestsController < ApplicationController
   before_action :fetch_requests, only: %i[show update edit]
 
-  def create
-    return unless params[:submit_req]
-
-    adds = add_requests(params[:ht_users])
-    flash[:notice] = "Added #{'request'.pluralize(adds.count)} for #{adds.join ', '}" if adds.count.positive?
+  def create # rubocop:disable Metrics/MethodLength
+    if params[:submit_requests]
+      @added_users = add_requests(params[:ht_users])
+      flash[:notice] = "Added #{'request'.pluralize(@added_users.count)} for #{@added_users.join ', '}" if @added_users.count.positive?
+      session[:added_users] = @added_users
+    end
+    if params[:submit_renewals]
+      @renewed_users = add_renewals(params[:ht_users])
+      flash[:notice] = "Renewed #{@renewed_users.join ', '}" if @renewed_users.count.positive?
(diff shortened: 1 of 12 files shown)


app/controllers/ht_approval_requests_controller.rb
```diff
@@ -3,18 +3,26 @@
...
+    end
     redirect_to action: :index
-    session[:added_users] = adds
   end
 
   def index
     @reqs = HTApprovalRequest.order('approver')
     @added_users = session[:added_users] || []
+    @renewed_users = session[:renewed_users] || []
+    session.delete :added_users
```
aelkiss, line 24: Looks like this was the fix for the stickiness?

aelkiss, line 24: Looks like this was the fix for the stickiness?

app/controllers/ht_approval_requests_controller.rb
```diff
@@ -3,18 +3,26 @@
...
 
-  def create
-    return unless params[:submit_req]
-
-    adds = add_requests(params[:ht_users])
-    flash[:notice] = "Added #{'request'.pluralize(adds.count)} for #{adds.join ', '}" if adds.count.positive?
+  def create # rubocop:disable Metrics/MethodLength
+    if params[:submit_requests]
+      @added_users = add_requests(params[:ht_users])
+      flash[:notice] = "Added #{'request'.pluralize(@added_users.count)} for #{@added_users.join ', '}" if @added_users.count.positive?
```
aelkiss, line 9: Consider moving into `add_requests` (and likewise below for the flash & session manipulation for `add_renewals`?)

aelkiss:
Going to go ahead and merge this, then will update ruby & rails

app/controllers/ht_approval_requests_controller.rb
moseshll: Yes, it was. I initially tried storing it in the flash, but the way our header deals with flash produced unwanted artifacts, so I went with this (I think flash is part of the session anyway, so six/half dozen).

app/controllers/ht_approval_requests_controller.rb
moseshll: On the list.

PR #54, opened 2020-05-22
Ht 2351 approval user email

moseshll:
Left open the option of previewing the HTML e-mail in the approval request edit page, but that would require instantiating HTUser in the view, or some other cleverness.

A couple of additional commits with routine fixes thrown in.


moseshll:
IIRC @base_url was needed with the preview at useradmin/rails/mailers

aelkiss:
Works for me without it..

moseshll:
Cool, prolly a testing artifact.

PR #57, opened 2020-05-26
HT-1195: make email to approvers editable

aelkiss:
- Adds ckeditor for HTML editing
- Does not (yet) persist the email anywhere after it's sent
- Disable turbolinks, since it messes with ckeditor


aelkiss:
@moseshll  I know you're on vacation -- we're trying to get this ready for Melissa to send emails on Monday, so I'm going to go ahead and merge it -- but please feel free to provide feedback when you're able and we can go back and revisit.

PR #62, opened 2020-06-02
HTS-139075: accept user_pid attribute for authentication

aelkiss:
This fixes an issue where users' authentication is rejected whose
identity provider do not send eduPersonPrincipalName or another
attribute that can be mapped to user_eid.

Otis::Authentication::UserPid is a straight copy-paste of
Keycard::Authentication::UserEid and included here without tests. It
should be tested when it is integrated into Keycard.

I will test this with useradmin-testing by removing Apache sending eduPersonPrincipalName, mail, etc headers and ensuring that authentication still succeeds.


aelkiss:
Thanks. I verified with useradmin-testing that if we pass the persistent ID header but not the eppn or mail attributes, it now passes authentication but (as expected) fails authorization for the main page. Merging.

PR #69, opened 2020-07-10
Ht 2456 checkpoint

moseshll:
I think at this stage the acceptance criteria are satisfied, even if the implementation is rough around the edges.

The new rake task `otis:migrate_users` will be *required* in Production in order to get Checkpoint grants for users in the Moku config. `Otis.config.staff` and `Otis.config.institution` (dreadful alternative to `staff_who_can_only_see_institutions`) config entries should also be added, though they are not required by the Rake task.


aelkiss:
One thing that might be nice to try to do is instead of having static grants set during testing to add them in the test setup. You can see how we did that in chipmunk, although a bunch is rspec-specific:

https://github.com/mlibrary/chipmunk/blob/master/spec/checkpoint_helper.rb
https://github.com/mlibrary/chipmunk/blob/master/spec/support/helpers/policy_helpers.rb

Not saying we need to do that now, but I do like how the policy specs came out in chipmunk, e.g:

https://github.com/mlibrary/chipmunk/blob/master/spec/policies/package_policy_spec.rb

moseshll:
ae338e3 at least runs the Rake task automatically. I don't think it's quite what you envisioned in that it assumes the same set of bogus users for all tests, but it does cut out a configuration step.

moseshll:
Attempted to send to staging (`ssh deployhost-001 deploy useradmin-staging HT-2456_checkpoint`) with progress ending as below,  no real smoking gun as to what failed:

```
Running `docker build` failed... examine /tmp/docker-build20200710-20908-10hasu4 for details.
Last five lines of output:
	/gems/ruby/2.6.0/gems/activesupport-5.2.4.3/lib/active_support/dependencies.rb:291:in `require'
	bin/rails:9:in `<main>'
	Tasks: TOP => environment
	(See full trace by running task with --trace)
	The command '/bin/sh -c bin/rails assets:precompile' returned a non-zero code: 1
```

aelkiss:
That's a little weird that `assets:precompile` would be the place it failed, but I'll see if I can figure out what's going on.

aelkiss:
This was a config issue; it's fixed now and deployed. I'll go ahead and merge it and deploy to production.

PR #73, opened 2020-07-20
HT-2477 Return in search should not submit form.

moseshll:


moseshll:
@billdueber I don't *think* I need to do anything about propagation. I am adding a note about it in the comments, however, in case there are issues in the future.

PR #77, opened 2020-08-03
HT-2468 one-time use links should not show menu links that are unauthorized

moseshll:
Construct the nav structure at a higher level than before and suppress the Home link and Go menu when the structure is empty.

### DIFF
--- a/app/views/shared/_header.html.erb
+++ b/app/views/shared/_header.html.erb
@@ -3,39 +3,29 @@
   <% unless logged_in? %>
       <p class="navbar-text">Not signed in</p>
   <% else %>
-
-    <% # Links menu will be empty when an external approver comes in.%>
-    <% # In that case we don't display any of the navigation bits. %>
-    <% nav_menu = [] %>
-    <% [['Users', 'users'],
-        ['Approval Requests', 'approval_requests'],
-        ['Institutions', 'institutions']].each do |item| %>
-      <% if can?(:index, "ht_#{item[1]}".to_sym) %>
-        <% nav_menu << [item[0], self.send("ht_#{item[1]}_path")] %>
-      <% end %>
-    <% end %>
-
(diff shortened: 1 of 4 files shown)


app/views/shared/_header.html.erb
```diff
@@ -1,33 +1,43 @@
...
-<% unless logged_in? %>
-    <p class="navbar-text">Not signed in</p>
-<% else %>
+  <% unless logged_in? %>
+      <p class="navbar-text">Not signed in</p>
+  <% else %>
+
+    <% # Links menu will be empty when an external approver comes in.%>
+    <% # In that case we don't display any of the navigation bits. %>
+    <% nav_menu = [] %>
```
aelkiss, line 9: This is the kind of thing that preferably would be pulled out into a presenter, but this logic should get pulled out into a presenter, but since this is shared I'm not sure exactly where it would go. Might be a good candidate for a top-level helper. I'll review a bit and see if I can come up with something here.

aelkiss: I think a helper is actually the way to go here, since (unlike with a presenter) this isn't really wrapping presentation of a specific AR model.

aelkiss: I suppose the controller could provide a `@nav_menu` object that was a fully-fledged thing, but that seems like overkill for the moment..

aelkiss:
Extracted as a helper and added a couple basic tests. If we find ourselves wanting to add more functionality to the navigation menu or further change the header we should probably pull out an object. That's a little bit difficult as it currently relies on the controller-level `can?` method, so we'd need to pass it the controller. More to think about for later as we add more/different kinds of functionality and roles in the application, but this works for now.

moseshll:
Looks good to me. I tried to write some simple tests, but got lost in the plumbing, so no go on that.

PR #83, opened 2020-11-24
Migrate from Travis to Github Actions

aelkiss:


aelkiss:
@moseshll  Looks like this is now passing with CI using GitHub actions instead of Travis. Any concerns or can we merge?

moseshll:
I don't have any concerns with this.

PR #84, opened 2021-02-05
HT-2155: Make institutions editable

aelkiss:
This adds support for adding and editing institution info.

After spending some time refactoring shib_admin it seemed more worthwhile to spend the time to move this functionality to OTIS, which is ultimately preferable anyway.

The proximal motivation for this was rethinking the workflow for setting up new institutions with respect to thinking about authentication vs. holdings -- in particular we wanted to add support for adding a new institution *without* having SAML metadata or authentication set up.

This also allows us to change ETAS status without needing to directly edit the database, which has been something we have kind of needed since last March... 

@mwarin  I think we could probably add support for display & editing of the institutions fields in `ht_billing_entities` pretty easily as well here.


aelkiss:
FYI @umjnega in case you'd like to review, or @rrotter as a replacement for the `ht_institutions_admin` functionality in shib_admin which you originally wrote.

aelkiss:
One thing this PR does not do is simplify the allowed affiliations -- it simply displays them as a regex without attempt to parse or validate. Future work could include better display & editing of the allowed affiliations and scopes.

app/presenters/ht_institution_presenter.rb
```diff
@@ -1,15 +1,140 @@
...
+  def login_test_url(eid = entityID)
+    "#{ht_base}/Shibboleth.sso/Login?entityID=#{eid}&target=#{ht_base}/cgi/whoami"
+  end
+
+  def mfa_test_url(eid = entityID)
+    "#{ht_base}/Shibboleth.sso/Login?entityID=#{eid}&authnContextClassRef=#{shib_authncontext_class}&target=#{ht_base}/cgi/whoami"
+  end
+
+  def metadata_url(eid = entityID)
+    "#{met_base}/met/entity/#{eid}"
```
aelkiss, line 138: We should consider pulling some or all of this out as configuration.

app/lib/saml_metadata.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+require 'nokogiri'
+require 'open-uri'
+
+class SAMLMetadata
+  MET_ENDPOINT = 'https://met.refeds.org/met/entity'
```
aelkiss, line 7: should pull this out as config

app/views/ht_institutions/index.html.erb
```diff
@@ -43,18 +43,44 @@
...
+        <td><%= i.metadata_link %></td>
       </tr>
     <% end %>
   </table>
+
+  <br />
+
+  <h2> Add New Institution </h2>
+
+  <div data-turbolinks="false">
```
aelkiss, line 64: should whack this div - was trying to debug a problem with the form below (it was missing `local:true`) that I was blaming on turbolinks

app/views/ht_institutions/show.html.erb
```diff
@@ -5,20 +5,34 @@
...
   <div class="col-sm-6">
     <dl class="dl-horizontal">
-      <dt>ID:</dt> <dd><%= @institution.inst_id %></dd>
-      <dt>GRIN Instance:</dt> <dd><%= @institution.grin_instance %></dd>
+      <dt>Institution ID:</dt> <dd><%= @institution.inst_id %></dd>
+      <dt>Domain:</dt> <dd> <%= @institution.domain %> </dd>
       <dt>Name:</dt> <dd><%= @institution.name %></dd>
+      <dt>Mapped Name:</dt> <dd>
+      <%= @institution.formatted_mapto_name %>
+      <!--      <p><small>Used when displaying the name of a logged-in user's institution, for example in the footer. This is typically used with university system members where individual campuses have separate log-in but the member is the university system as a whole.</small></p> -->
```
aelkiss, line 13: Might be nice to have this documentation text available; not sure the view is really the right spot for it; not sure how to best format it.

test/fixtures/umich_metadata.xml
```diff
@@ -0,0 +1,117 @@
+<md:EntityDescriptor xmlns:shibmd="urn:mace:shibboleth:metadata:1.0" xmlns:remd="http://refeds.org/metadata" xmlns:md="urn:oasis:names:tc:SAML:2.0:metadata" xmlns:ds="http://www.w3.org/2000/09/xmldsig#" xmlns:mdattr="urn:oasis:names:tc:SAML:metadata:attribute" xmlns:mdrpi="urn:oasis:names:tc:SAML:metadata:rpi" xmlns:mdui="urn:oasis:names:tc:SAML:metadata:ui" xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion" xmlns:xrd="http://docs.oasis-open.org/ns/xri/xrd-1.0" entityID="https://shibboleth.umich.edu/idp/shibboleth">
```
aelkiss, line 1: Should perhaps consider creating true faked-up metadata instead of just copying umich's metadata. OTOH, that metadata is public anyway, so ¯\_(ツ)_/¯

aelkiss:
There is also a bug where checking the 'US' checkbox does not actually update it.

app/controllers/ht_institutions_controller.rb
```diff
@@ -1,15 +1,65 @@
 # frozen_string_literal: true
 
 class HTInstitutionsController < ApplicationController
-  before_action :fetch_institution, only: %i[show]
+  before_action :fetch_institution, only: %i[show edit update]
+
+  PERMITTED_UPDATE_FIELDS = %i[
+    grin_instance
+    name
```
moseshll, line 8: The bug with the US checkbox not doing anything is solved by adding `us` to this list. I note that once changed to from non-US to US, can't change it back in the UI. Is this by design?

aelkiss: Probably not -- probably makes sense just to have it be fully editable. I think my initial thought was that it wasn't something that would change.

app/views/ht_institutions/show.html.erb
moseshll: I tried it as a Bootstrap tooltip, and it looks OK-ish. Maybe that's for a different iteration.

```
<script>
$(function () {
  $('[data-toggle="tooltip"]').tooltip()
})
</script>
...
<a href="#" data-toggle="tooltip" title="Used when displaying the name of
          a logged-in user's institution, for example in the footer.
          This is typically used with university system members where individual
          campuses have separate log-in but the member is the university system
          as a whole.">
        <i class="glyphicon glyphicon-info-sign"></i>
      </a>
```

moseshll (approved):

app/controllers/ht_institutions_controller.rb
moseshll: Institutions seceding from the Union... hope not :-/
But, since add institution functionality is emerging, makes sense to treat the edit form like one would a fully editable new institution.

aelkiss:
> There is also a bug where the 'US' checkbox appears to be editable for existing non-US institutions, but checking it does not do anything.

This should be resolved now.

app/views/ht_institutions/show.html.erb
aelkiss: I tried experimenting with some stuff and just made it worse. Worth re-thinking later how to provide help (both here and for users)

aelkiss:
@moseshll  Ready for re-review

app/lib/saml_metadata.rb
```diff
@@ -0,0 +1,34 @@
...
+    @entity_id = entity_id
+    @doc = Nokogiri::XML(data)
+  end
+
+  def name
+    @doc.xpath('//mdui:DisplayName|//md:organizationDisplayName|//md:OrganizationName').first.text
+  end
+
+  def scopes
+    @doc.xpath('//md:IDPSSODescriptor/md:Extensions/shibmd:Scope').map(&:text)
```
moseshll, line 24: Seeing Nokogiri::XML::XPath::SyntaxError in HTInstitutionsController#new when I drop in various randomly-selected entity IDs here. For example https://shibboleth.fhwn.ac.at/idp/shibboleth

moseshll (changes requested):
See comment on SAMLMetadata#scopes errors

app/lib/saml_metadata.rb
aelkiss: I'll take a look -- thanks!

aelkiss: Unfortunately met.refeds.org seems to be down right now, but I suspect this is a namespace issue and that I know how to fix it.

aelkiss: I pushed a fix that should address this.

app/views/ht_institutions/show.html.erb
```diff
@@ -1,28 +1,35 @@
...
+})
+</script>
+
 <div id="maincontent" class="row">
   <h1><%= @institution.name %></h1>
 
   <%= render 'shared/flash_message' %>
 
-  <div class="col-sm-6">
-    <dl class="dl-horizontal">
```
moseshll, line 7: Removing dl-horizontal seems to result in stacking the dt tags on top of the dd tags, which I don't think is the aim (I can attach a screenshot if we're seeing something different layout-wise). Additionally, when I run HTML validation it looks like the divs inside the dl make the parser unhappy.

moseshll (changes requested):
See single comment about layout changes.

app/views/ht_institutions/show.html.erb
aelkiss: I think I committed something I didn't intend to. Thanks for bearing with me...

aelkiss: Force pushed a corrected commit; this should be fixed now

moseshll (approved):

PR #86, opened 2021-02-23
Use MariaDB for dev & test; add editing for billing members fields

aelkiss:
@mwarin Would be helpful just to verify I've done sufficient here for adding & editing the billing fields.

@moseshll `HTInstitutionsController` now handles accepting the embedded fields for the "billing member" - I don't love the way the parameter handling happens there, but I'm not sure there's really a better way. Open to suggestions.

@botimer I primarily wanted to point out the other commit to you (using MariaDB for dev & test database) to point out a couple things I'm not totally happy with with respect to keycard/checkpoint/docker -- in particular, I would prefer to use a one-shot container to run the `checkpoint:migrate`, `keycard:migrate`, `db:seed`, etc, however, if I use the pattern from https://learning.oreilly.com/library/view/docker-for-rails/9781680506730/f_0104.xhtml there doesn't seem to be a way to ensure that completes before the web container starts -- but while normal Rails apps are (normally, so far as I recall) OK if they start up before the `db:migrate` runs, with Checkpoint they check on initialization if the `grants` table is there. I suppose I could try to add a script that waits for that table to be present before starting the app, but... ugh. I suppose I could run the `db:seed` in a one-shot container and run the `checkpoint:migrate` and `keycard:migrate` every time before starting the web container, but that (conceptually) doesn't seem quite right either. Let me know if you have any thoughts out of that conundrum. Also perhaps worth thinking if you have a good pattern for orchestrating those sorts of things in Kubernetes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,14 +18,11 @@ jobs:
       - name: Set up tests
         run: |
           docker-compose build
-          docker-compose run web bundle install
-          docker-compose run web bundle exec rake keycard:migrate RAILS_ENV=test
-          docker-compose run web bundle exec rake checkpoint:migrate RAILS_ENV=test
-          docker-compose run -e RAILS_ENV=test web bundle exec rails db:setup
+          docker-compose run test bundle install
 
       - name: Run rubocop
-        run: docker-compose run web bundle exec rubocop
+        run: docker-compose run test bundle exec rubocop
 
       - name: Run tests
-        run: docker-compose run web bundle exec rails test
+        run: docker-compose run test
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,17 +4,18 @@ ARG UID=1000
 ARG GID=1000
 
 RUN apt-get update -yqq && apt-get install -yqq --no-install-recommends \
-  nodejs
+  nodejs \
+  netcat
 
 # COPY Gemfile* /usr/src/app/
 WORKDIR /usr/src/app
-# 
+#
 ENV BUNDLE_PATH /gems
-# 
+#
 RUN gem install bundler
-# 
+#
 # COPY . /usr/src/app
 
-CMD ["bin/rails", "s", "-b", "0.0.0.0"]
+CMD ["bin/rails","s","-b","0.0.0.0"]
 
 
--- a/README.md
+++ b/README.md
@@ -1,48 +1,32 @@
-# Otis Elevated Access Tool
+# Otis Administrative Tools
 
 ## Initial setup
-### 1. Getting a local copy, bundle install gems, and execute setup script
+### 1. Set up development
 
 ```
 $ git clone https://github.com/hathitrust/otis.git
 $ cd otis
-$ bundle install
+$ docker-compose build
+$ docker-compose run web bundle install
 ```
 
-### 2. Database
-
-Development mode uses sqlite3 with generated data. The keycard and
-checkpoint databases also need to be set up. The `otis:migrate_users` task
-sets up three dummy users with different levels of access (this is done
-automatically for the `test` environment by the test script).
+### 2. Trying it out
 
 ```
-bundle exec rake keycard:migrate RAILS_ENV=development
-bundle exec rake keycard:migrate RAILS_ENV=test
-bundle exec rake checkpoint:migrate RAILS_ENV=development
-bundle exec rake checkpoint:migrate RAILS_ENV=test
-bundle exec rake db:setup
-bundle exec rake otis:migrate_users RAILS_ENV=development
+docker-compose up -d web
 ```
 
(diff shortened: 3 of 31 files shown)


botimer:
Thanks, @aelkiss. I was just working up a Dockerfile / compose setup for another project and the Keycard/Checkpoint setup was really rough. I need to return to this and see how we can improve the ergonomics.  I'm aware of [dockerize](https://github.com/jwilder/dockerize), but I don't think it solves sequencing of one-shot things as much as "wait for readiness". I also ran into problems with spring/bootsnap vs. the initialization order.

So, some part of this is needed convenience/robustness on the gem side, and some is orchestration of containers -- you are definitely not the only people needing this.

moseshll:
Are we concerned with validating the added fields? OCLC/MARC look too hard to fully validate (alphabetic check?), country codes maybe, how about weight? Is that always between 0.0-1.0 inclusive? I entered some alphabetic gibberish and it failed silently on update, just preserved to existing value.

As an aside, I note a nonfatal error on startup with `db/seeds.rb` apparently causing `Mysql2::Error: Field 'id' doesn't have a default value: INSERT INTO ht_approval_requests …`. I guess there's an `id` primary key under the hood there, so because it's fake data I just modded the initializer with…
```
ar = HTApprovalRequest.new(
    id: Faker::Number.unique.number(digits: 9),
    approver: user.approver,
    userid: user.email,
    received: received,
    renewed: renewed
  )
```

I tweaked `seeds.rb` so I could have something to look at...

```
def create_ht_billing_member(inst_id)
  HTBillingMember.create(
    inst_id: inst_id,
    weight: Faker::Number.within(range: 0.0..1.0),
    oclc_sym: Faker::Alphanumeric.alpha(number: Faker::Number.between(from: 3, to: 5)),
    marc21_sym: Faker::Alphanumeric.alpha(number: 3).upcase,
    country_code: Faker::Address.country_code,
    status: 1
  )
end

10.times do
  inst_id = create_ht_institution(1)
  create_ht_billing_member(inst_id) if [0, 1].sample.zero?
end
```

aelkiss:
Weight can be greater than 1. @mwarin  should we validate for any particular set? I agree validating the OCLC or MARC codes is probably not that useful -- @mwarin  any thoughts there?

@moseshll  Can you push a commit to this PR with the updates above when you have a chance?

aelkiss:
Pushed a commit to show last update date for institutions.

moseshll:
Two variations on `weight` validation I was noodling with:

```
# frozen_string_literal: true

class HTBillingMember < ApplicationRecord
  belongs_to :ht_institution, foreign_key: 'inst_id', optional: true

  self.primary_key = 'inst_id'
  
  #validates_numericality_of :weight, greater_than_or_equal_to: 0,
  #                          less_than: 100, only_integer: false,
  #                          allow_nil: true
  validates :weight, format: { with: /\A\d+(?:\.\d{0,2})?\z/ },
                     numericality: { greater_than_or_equal_to: 0,
                                     less_than: 100 },
                     allow_nil: true
end
```

botimer:
I would be tempted to try a custom type. This isn't the most complicated validation, nor is a weight with 0-100 range likely to gain a lot business logic, but it's nice to have a value object to deal with basic casting/serialization. Then you have somewhere to put logic (and test it easily) if you need it, rather than the business rules only existing as field-level DSL stuff. This looks like a simple place to practice with custom types and see what you think of less "stringly typed" attributes.

Put more simply: I think Weight is a domain concept here and Rails' Attributes API gives you a way to use an object easily, instead of a specially formatted/ranged primitive.

Some examples:

https://bigbinary.com/blog/rails-5-attributes-api
https://aliou.me/posts/2019/10/attributes-api-and-value-objects/

aelkiss:
Pushed another commit to add audit logging for changes to `HTInstitution` - useful if for no other reason so we can go back and check when we enabled or disabled ETAS for an institution and with what affiliations. (Maybe would have been more useful if we had it in retrospect, but still useful going forwards.)

aelkiss:
@mwarin I think it probably needs you to weigh in on validation and/or custom type for weight. My sense is the system members complicate things, but you have a better understanding of that than me.

aelkiss:
@mwarin says that we should probably leave the weight as a free-form float for the time being.

aelkiss:
There's code in the catalog indexing that hits LoC to pull language mappings, etc, so it probably wouldn't be hard to pull down institutional mappings & make sure they stay up to date, but that feels like scope creep. There is not currently any validation for the OCLC or MARC code, and the consequences of bad data there are minimal.

aelkiss:
Based on discussion with @mwarin we're going to merge this as-is for now. We may decide to come back later and add more functionality around validations for the new fields, especially when we want to add more functionality around reporting in a front-end.

PR #87, opened 2021-03-10
db schema updates; billing member fixes

aelkiss:
- Remove `ht_institutions` fields that are no longer present in production

- Improve behavior for billing members. The `accepts_attributes_for` stuff is kind of a mess, esp. in figuring out whether or not to add/edit something; longer term, it would probably be better to explicitly just add a form for that object rather than having it as nested attributes. I thought `accepts_attributes_for` would be easier since it's a 1:1 relationship, but in fact it creates a bit of a mess. I think if we want to make more changes to billing members it's probably worth experimenting with other options here.


aelkiss:
@mwarin  take a look and see if the new behavior here makes sense; @moseshll  you may want to take a look and see if there's anything we could do to make this clearer in the short term

moseshll (approved):

PR #88, opened 2021-03-22
Remove reliance on moku for deployment

aelkiss:
This adds tanka & jsonnet config, and changes the app config enough so that it can 1) build a production image and 2) use config from k8s instead of 

I'm particularly interested in your thoughts on production.yml + env vars vs. the configmap - frustratingly, we can't just dump production.local.yml in via a configmap, AFAICT. I guess another possibility would be referring to env vars in production.yml + using dotenv + a configmap on .env -- that might be a little bit cleaner than this at the expense of slightly more indirection. Let me know what you think.

Marking this as draft for now; I still need to add the github actions for deployment.


kube/lib/otis/config.libsonnet
```diff
@@ -0,0 +1,29 @@
...
+          configmap: 'production-config',
+          # path to mount above configmap
+          path: '/usr/src/app/config/settings/production',
+          # key from configmap to use for application config
+          key: 'otis.yml'
+        }
+      },
+      database: {
+        port: 3306,
+        ip: '10.255.8.249',
```
aelkiss, line 19: I'm not wild about having this here, but I guess it's at least short-lived until consul is available.

botimer:
Hmmm... I have been thinking about how best to manage instance-specific configuration. I think part of the issue is that there are so many viable options and subtle differences that don't add up to much.

With config/ettin in place, it's pretty easy to have a settings.local.yml or production.local.yml with whatever specific values you need, without clobbering the general stuff in settings.yml or production.yml. You could definitely drop in one of those with a ConfigMap / volumeMount. Likewise, you could pick up a .env file or something, but picking and mapping individual keys through at least two layers of config is pretty gross. I personally favor using settings.local.yml and only exposing a couple of fundamental things for ENV override (unless there is a more automated way to declare/set the options and precedence I haven't seen).

As the authoritative home for the config/files, I'm less settled. I like the proximity and directness of ConfigMaps (whether for env vars or file contents), but they are pretty clunky to work with, and effectively runtime. What I am pretty certain of is that "shape" things are fine to have in an app repo, but value things should be elsewhere. What I am less certain of where exactly they should be.

It's pretty clear that "gitops" is taking over in general, and Flux is a leading player. What is way less clear is how Flux and Tanka will interact. The team working on Flux 2 (which is nearly to GA status) has made a lot of provision for Helm and Kustomize, but I'm not sure how the Flux / Tanka teams see each other.

Argo is in the mix, too. It is more involved, but has some direct provisions for shared (multi-tenant) clusters and multi-cluster deployment. I'm leery of (Capital-E) Enterprise complexity at this stage and really want to grow into what we need. The recipe needs to stay pretty simple for simple needs. Ideally, the recipes can scale up for more complex needs, rather than be completely different approaches. But there is much to do in figuring this out...

aelkiss:
The problem is that it's a bit clunky as Ettin stands now (see the workaround in the PR), because it's expecting that `settings.local.yml` or `production.local.yml`  will be in the same directory as the application-default `production.yml` -- but if we use a configmap mounted in `config/settings`, that hides from view anything in the same directory. That said, it sounds like you would lean towards what's here now rather than additional layers of mapping via the environment.

Not sure if the question of Tanka vs. something else is relevant to the question of runtime configuration, or if it's just a more general observation. I haven't done any research in this area, and I think there is a strong reason to follow A&E's lead here. Mainly, I just need to stay up to date on whatever A&E's best practices are, and put some work into understanding them (which to some first approximation I do now for Tanka)

botimer:
Oh oh oh! You can mount a single file from a ConfigMap without wiping the directory. I ran into that before; sorry I didn't mention it. Doing single files depends on `mountPath` and `subPath`.

https://carlos.mendible.com/2019/02/10/kubernetes-mount-file-pod-with-configmap/
https://medium.com/hackernoon/mount-file-to-kubernetes-pod-without-deleting-the-existing-file-in-the-docker-container-in-the-88b5d11661a6
https://kubernetes.io/docs/concepts/storage/volumes/#using-subpath

It's a real hassle if you have a _lot_ of files to map into a directory you can't replace, but one or two isn't bad.

aelkiss:
Thanks -- I removed the `OTIS_CONFIG` stuff. Going to go ahead and merge this to ensure the actions are all working as expected, but please feel free to comment further.

aelkiss:
I will probably add gem & docker layer caching here; this might be a good point to look at shared github actions between this,  hathi_search_client, and patron_account

botimer:
Yes, we have been thinking about shared actions, but for the moment, we are consciously letting the concrete cases emerge. This is based on the hard-won experience that copypasta and early extraction are nearly equivalent as far as maintenance and understanding. We researched how other teams are managing actions across repos, and we find significant duplication. That is not to say that duplicating swathes of yaml is the best way, but that it is certainly common.

What we're trying to go for is a small set of utility items that can be reused where needed. We don't want one configurable mega action, nor 100 to shop through. The initial instinct is that we have 3 or 4 patterns for the tricky/idiosyncratic stuff that can move to a library, and everything else should be inlined.

PR #89, opened 2021-03-30
Configuration for production & staging

aelkiss:


aelkiss:
FYI @daaang - copies actions from https://github.com/mlibrary/mattlach-phineas-testing-actions

PR #90, opened 2021-04-05
HT-1460 Institutions table: show how many people with access by insti…

moseshll:
…tution

Opting for the simpler approach of displaying this info on the institution show page.

### DIFF
--- a/app/presenters/ht_institution_presenter.rb
+++ b/app/presenters/ht_institution_presenter.rb
@@ -102,6 +102,14 @@ class HTInstitutionPresenter < SimpleDelegator
     end
   end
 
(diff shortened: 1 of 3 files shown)


aelkiss (approved):
Nice. Do we want to also actually show the users (or have a link to the user page with an appropriate filter?)

moseshll:
My feeling is that this is meant to be zero/not-zero sanity check on the counts. Given that the index page is sortable and searchable by institution I'd say the information need not be further elaborated upon (for now).

aelkiss:
:+1:

aelkiss:
:+1:

PR #93, opened 2021-04-07
HT-2826

moseshll:
Fix OTIS WCAG contrast issues and HTML validation
- Badge and link text contrast
- Label-less checkboxes (I think the changes here are real UI improvements)
- Missing <th> text content
- Missing Bootstrap icon fonts
- Obvious stray anchors in nav menu
- Minor HTML errors: illegal `<hr>` placement, stray end tags


app/views/ht_approval_requests/index.html.erb
```diff
@@ -24,23 +25,23 @@
...
-                can?(:edit, :ht_approval_requests) %>
+          <% if can?(:edit, :ht_approval_requests) %>
             <%= link_to req.approver, edit_ht_approval_request_path(req.approver) %>
           <% else %>
             <%= req.approver %>
           <% end %>
         </td>
-        <td><%= link_to req.userid, ht_user_path(req.userid) %></td>
+        <td>
+          <% if can?(:edit, :ht_approval_requests) %>
```
aelkiss, line 36: I'd suggest extracting this logic to `HTUserPresenter`

app/views/ht_users/edit.html.erb
```diff
@@ -64,21 +62,24 @@
...
         <dt>Renewal Status</dt>
         <dd><%= @user.badge %></dd>
-        <dt><%= form.label :iprestrict %></dt>
+        <dt><%= form.label :iprestrict, for: :iprestrict_field %></dt>
         <dd><%= form.text_field :iprestrict, value: @user.iprestrict&.join(', '), size: 40, id: :iprestrict_field, disabled: @user.mfa %>
           <p class="text-muted"><%= t('ht_user.edit.iprestrict_prompt') %></p>
         </dd>
-        <dt>Multi-Factor?:</dt>
         <% if @user.ht_institution.shib_authncontext_class.present? %>
+          <dt><%= form.label 'Multi-Factor?:', for: :mfa_checkbox %></dt>
```
aelkiss, line 70: Also worth pulling out this logic to a presenter?

app/views/ht_approval_requests/index.html.erb
aelkiss: Can you explain why the validator required avoiding a standalone checkbox column? I definitely prefer the way it looked before; I don't think this is worth spending a lot of time on, but I at least want to understand what the issue was.

app/views/ht_institutions/_form.html.erb
```diff
@@ -50,25 +49,24 @@
           <%= check_box_tag :create_billing_member %>
```
aelkiss, line 49: This should be pulled down into the other `<dl>`

aelkiss (commented):
Generally looks good; there are a couple of places I think it would be worthwhile to extract logic from the views to the presenter, and I also want to understand what the issue was that led to putting the checkbox in the email column.

app/views/ht_approval_requests/index.html.erb
moseshll: The error I get is "form elements must be labeled" (this is using the Firefox dev tools accessibility checker). Using the ht_user e-mail as a label seems better than taking up real estate with a do-nothing label like "Select: " in each row. I can use the e-mail field as a label across table cells, if it's preferable to keep a dedicated "Select" column.

And I agree that the logic belongs in the presenter.

app/views/ht_users/edit.html.erb
moseshll: While I'm at it...

app/views/ht_approval_requests/index.html.erb
aelkiss: I see -- yeah, I agree using the e-mail as a label makes sense; I think it would be preferable to use it as a label across table cells if possible.

moseshll: It does work to go cross-cell for labeling. To my surprise, frankly.

test/controllers/ht_approval_requests_controller_test.rb
```diff
@@ -23,6 +23,7 @@ def setup
     get ht_approval_requests_url
     validator = W3CValidators::NuValidator.new
     w3c_errs = validator.validate_text(@response.body).errors
+    sleep 1
```
aelkiss, line 26: This is because it's sending it to an online service because it's rate-limiting us? If so, perhaps we should create an issue for running our own copy when running tests -- e.g. should be easy to spin up https://hub.docker.com/r/validator/validator/ via docker-compose when running tests

moseshll: Yes. I didn't see a timeout until after the second commit on the PR resulting in the tests failing. I saw it happen once locally, following that (could also have been a transient, who knows). I noted that the validator gem's author mentions a `sleep 1` call in that repo's tests, so just went with that for the time being.

aelkiss: :+1:

aelkiss (approved):

aelkiss:
Changes to view logic / presenters look good. I'll go ahead and merge.

aelkiss:
Once this is (automatically) deployed to testing (https://beta-1.babel.hathitrust.org/otis-testing) and we've verified it looks OK, go ahead and deploy to production -- draft a release in GitHub, then use the action to deploy that release to production (https://github.com/hathitrust/otis/actions/workflows/deploy-production.yaml)

PR #94, opened 2021-04-08
HT-2413: Set Reply-To from configuration

aelkiss:
I've already set this in the production config, so it should get picked up as soon as we deploy this.


moseshll (approved):
Looks good here.

PR #95, opened 2021-04-12
Run assets:precompile at runtime after RAILS_RELATIVE_URL_ROOT is set

moseshll:
Not sure about changing the `CMD` in `Dockerfile` as opposed to `Dockerfile.prod` -- I believe it to be unused. Changing it to parallel the production file anyway.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,6 +16,4 @@ RUN gem install bundler
 #
 # COPY . /usr/src/app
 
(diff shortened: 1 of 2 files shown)


aelkiss:
The regular `Dockerfile` should (normally) only run in dev mode so I don't think we need it there

`Dockerfile.prod` is used here and is what gets pushed to docker hub: https://github.com/hathitrust/otis/blob/master/.github/workflows/build-master.yaml#L27

PR #96, opened 2021-04-28
HT-2403 'approved' renewal requests expire so they can be renewed aga…

moseshll:
…in next year

Approach this like dividing users into active and inactive, with requests that have been renewed sequestered in a second list. A user with a renewed (the ticket says "approved" but I think that is an oversight) request should no longer be prevented from having a new approval request created.

It may be in future the inactive/renewed requests need not even be acknowledged in the UI. Let's go with this for now. I'd rather hide them than do a mass deletion.

No attempt has been made to be smart about when to allow a new request to be generated. I trust the "Expires" column will be the primary driver.

### DIFF
--- a/app/controllers/ht_approval_requests_controller.rb
+++ b/app/controllers/ht_approval_requests_controller.rb
@@ -18,8 +18,9 @@ class HTApprovalRequestsController < ApplicationController
   end
 
   def index
-    reqs = HTApprovalRequest.order('approver')
-    @reqs = reqs.map { |r| HTApprovalRequestPresenter.new(r) }
+    requests = HTApprovalRequest.order('approver')
+    @active_reqs = requests.not_renewed.map { |r| HTApprovalRequestPresenter.new(r) }
+    @inactive_reqs = requests.renewed.map { |r| HTApprovalRequestPresenter.new(r) }
     @added_users = session[:added_users] || []
     @renewed_users = session[:renewed_users] || []
     session.delete :added_users
@@ -61,7 +62,7 @@ class HTApprovalRequestsController < ApplicationController
   end
 
   # Add an approval request for selected users.
-  # If one already exists, silently skip over it.
+  # If unrenewed one already exists, silently skip over it.
   # Returns an Array of ht_user emails added to requests.
   def add_requests(emails) # rubocop:disable Metrics/MethodLength
     if emails.nil? || emails.empty?
@@ -70,7 +71,7 @@ class HTApprovalRequestsController < ApplicationController
     end
     adds = []
     emails.each do |e|
-      next if HTApprovalRequest.where(userid: e).count.positive?
+      next if HTApprovalRequest.where(userid: e, renewed: nil).count.positive?
 
       begin
         add_request e
--- a/app/controllers/ht_approval_requests_controller.rb
+++ b/app/controllers/ht_approval_requests_controller.rb
@@ -19,8 +19,8 @@ class HTApprovalRequestsController < ApplicationController
 
   def index
     requests = HTApprovalRequest.order('approver')
-    @active_reqs = requests.not_renewed.map { |r| HTApprovalRequestPresenter.new(r) }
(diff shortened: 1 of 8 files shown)


app/controllers/ht_approval_requests_controller.rb
```diff
@@ -18,8 +18,9 @@ def create # rubocop:disable Metrics/MethodLength
   end
 
   def index
-    reqs = HTApprovalRequest.order('approver')
-    @reqs = reqs.map { |r| HTApprovalRequestPresenter.new(r) }
+    requests = HTApprovalRequest.order('approver')
+    @active_reqs = requests.not_renewed.map { |r| HTApprovalRequestPresenter.new(r) }
+    @inactive_reqs = requests.renewed.map { |r| HTApprovalRequestPresenter.new(r) }
```
aelkiss, line 23: I wonder if "completed" might be a better term?

test/models/ht_approval_request_test.rb
```diff
@@ -45,6 +45,22 @@ class HTApprovalRequestTest < ActiveSupport::TestCase
...
     assert_equal(HTApprovalRequest.digest(token), req.token_hash)
   end
+
+  test 'has correct Checkpoint attributes' do
+    req = create(:ht_approval_request, :expired)
+    assert_equal :ht_approval_request, req.resource_type
+    assert_equal req.id, req.resource_id
+  end
+
+  test 'for_user finds most recent or most incomplete' do
```
aelkiss, line 55: How/where are we relying on this behavior? It's a little bit surprising to me for this behavior to be included with `for_user` -- something like this might better expose the intention:

```ruby
def self.most_recent(user)
  for_user(user).order(...)
end
```

aelkiss (commented):
I tried this out locally and as far as I can tell the behavior is what we're expecting. I had a couple comments inline.

app/controllers/ht_approval_requests_controller.rb
moseshll: I'll meet you half way with "complete" and "incomplete"

aelkiss: :+1:

test/models/ht_approval_request_test.rb
moseshll: This is the first time we've explicitly allowed multiple requests for a single user, so this is grabbing the most recent and/or most NULL-heavy) ones.

app/models/ht_approval_request.rb
```diff
@@ -2,9 +2,18 @@
 
 class HTApprovalRequest < ApplicationRecord
   self.primary_key = 'id'
+  # "Active" requests that may be subject to further action
   scope :not_renewed, -> { where(renewed: nil) }
+  # "Inactive" requests that are only of historical interest
+  scope :renewed, -> { where.not(renewed: nil) }
   scope :for_approver, ->(approver) { where(approver: approver).order(:sent, :received, :renewed) }
-  scope :for_user, ->(user) { where(userid: user) }
+  scope :for_user, ->(user) { where(userid: user).order(:sent, :received, :renewed) }
```
aelkiss, line 10: It looks like `for_user` is only currently used in `most_recent` below, so I'm not sure we need the order here.

aelkiss (approved):
I think this is more clear to me now. I had one small comment about the extra order on `for_user`, but I don't think that needs to block merging this.

app/models/ht_approval_request.rb
moseshll: I'd rather have the order be explicit and mirroring  `for_approver` rather than sorta random, even if it is never used.

moseshll:
I'll plan to merge and make sure the deployment to test.babel works on Thu prior to doing the tag/release dance.

aelkiss:
Sounds good, thanks!

app/models/ht_approval_request.rb
aelkiss: 👍

PR #98, opened 2021-04-30
Addresses CVE-2021-28965 vulnerability for rexml gem < 3.2.5

moseshll:


respinos:
Are you using rexml anywhere?

aelkiss:
I'd be hesitant about adding it to the `Gemfile` - does `bundle update rexml` do the trick? I doubt we are using it anywhere explicitly, but it's good to keep on top of these things especially when they don't involve breaking changes -- and luckily we have pretty good coverage for this project, so we can be confident of updating.

moseshll:
@respinos No, we're not using it directly. As far as I can tell just doing `bundle update rexml` should work -- I tried doing something similar in GitHub earlier but it broke the tests. I'll try just dropping the new version into `Gemfile.lock` and see if it does the trick on a different branch.

moseshll:
Heh, never mind. I see dependabot did my work for me. Closing this PR.

PR #101, opened 2021-05-06
Update gems; Use standard.rb instead of rubocop

aelkiss:
standardrb: https://github.com/testdouble/standard

I noticed when I updated rails that a number of other gems were quite outdated, in particular rubocop. Rather than just updating rubocop to the latest I thought it might be worthwhile trying out standardrb. The vast majority of the changes it makes below have to do with single vs. double quotes; the others changes don't seem particularly offensive even if they aren't always the choice I would necessarily make.


moseshll (approved):
It will be interesting to experiment with a less pedantic RuboCop alternative. The preference for double quotes will trip me up a lot. The leading dot chains will be interesting, I have irked RuboCop in various ways when trying to creatively chain methods/blocks in other projects. I'm all-in.

PR #104, opened 2021-06-22
use RAILS_LOG_LEVEL; log at INFO level by default in production

aelkiss:
Rails 5 defaulted to debug logging in production, but this is noisy, can leak PII, and not allowed by A&E policy.

This changes the default level to info but allows overriding it via `RAILS_LOG_LEVEL`, then sets it back to debug for testing & staging.

FYI @botimer


moseshll (approved):
Looks good. Glad to see this.

PR #106, opened 2021-08-09
HT-2452 do not re-request approval for expired users.

moseshll:
Applies only to manually-expired users, when the approval request has not been sent or responded to by approver.
Works in the same way as manual change of approver, by deleting stale approval requests.

### DIFF
--- a/app/models/ht_user.rb
+++ b/app/models/ht_user.rb
@@ -179,6 +179,9 @@ class HTUser < ApplicationRecord
   private
 
   def clean_requests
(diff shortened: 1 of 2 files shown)


aelkiss (approved):
Code & tests look good to me. I tested this scenario locally and it works as described:

1) create approval request & send email for approver for multiple users
2) expire one of the users manually
3) verify that the expired user's approval request has been deleted
4) expire the approvers' other approval requests (manually from rails console)
5) verify that the expired user does not show up when re-sending the request

PR #108, opened 2021-08-17
Adds HTContact and HTContactType models.

moseshll:
- Adds creation, editing, deletion to users with write access to the two tables.
- Removes direct support for ETAS contact as ht_institution field.
- Adds 'contact' user role.

Caveats:
- I had to hack around a lot with FactoryBot ht_contact factory and I don't think it is very nice.
- ~~db-image needs to be updated once the proposed schema is finalized.~~

### DIFF
--- a/lib/tasks/migrate_users.rake
+++ b/lib/tasks/migrate_users.rake
@@ -16,6 +16,8 @@ namespace :otis do
     admin_role = Checkpoint::Credential::Role.new(:admin)
     view_role = Checkpoint::Credential::Role.new(:view)
     res_wildcard = Checkpoint::Resource::AllOfAnyType.new
+    res_contact = Checkpoint::Resource::AllOfType.new(:ht_contacts)
+    res_contact_type = Checkpoint::Resource::AllOfType.new(:ht_contact_types)
     if Otis.config.users.present?
       Otis.config.users.each do |u|
         agent = Checkpoint::Agent::Token.new("user", u)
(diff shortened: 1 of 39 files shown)


lib/tasks/migrate_users.rake
```diff
@@ -39,6 +41,26 @@ namespace :otis do
...
         end
+        unless Services.checkpoint.permits?(agent, view_role, res_contact)
+          Services.checkpoint.grant!(agent, view_role, res_contact)
+        end
+        unless Services.checkpoint.permits?(agent, view_role, res_contact_type)
+          Services.checkpoint.grant!(agent, view_role, res_contact_type)
+        end
+      end
+    end
+    if Otis.config.contact.present?
```
aelkiss, line 52: The names of the keys in `Otis.config` for permissions have become a little bit confusing. Essentially there should be four roles now:

- superuser (should be granted admin all)
- staff (should be granted view all)
- institution admin (should be granted admin institutions)
- contact admin (should be granted admin contacts, admin contact type)

We may want to start making these roles more explicit in checkpoint but I would need to think about exactly how we do that.

aelkiss:
I took a look through and I didn't see any red flags. I will try it out in development.

I've run into similar issues with factories & associations in the past. If it works, it works; the tests aren't hideously slow (33 sec for everything on my machine, and I suspect a lot of that might be the external W3C validator?)

lib/tasks/migrate_users.rake
moseshll: I looked into "policy" objects in the Chipmunk repo (and also in Heliotrope but it's a lot more complicated) and I agree we are pushing up against maintainability limits in the way we do roles. But, decided it was enough out of scope for this round that can-kicking-down-the-road was called for.

moseshll:
It may be worth conditioning W3C checks on an environment variable, with the default being OFF.

lib/tasks/migrate_users.rake
aelkiss: :+1:

aelkiss (approved):
I tried in development and this all looks fine. I think we will want to create the new contacts & contacts type table in production before deploying to production. Let's wait until Tuesday for that.

aelkiss:
We will probably also want:

- the ability to display all contacts of a given type for all institutions
- the ability to download that info

Those are follow-up issues, though.

PR #111, opened 2021-09-01
HT-1451 Audit Log for OTIS

moseshll:
- Merge `HTInstitutionLog` and `HTUserLog` into new class `HTLog`
- Replace `ht_institution_logs` and `ht_user_logs` tables with `ht_log`
- Add event logging to `HTContact` and `HTContactType`
- Add logs index page with JSON download
- Caveats:
  - Deletion logging for contacts and contact types is done before checking success so object can be logged.
  - `ht_logs.model` is derived from Keycard `#resource_type` rather than `#class` because presenters muddy the waters.


app/controllers/application_controller.rb
```diff
@@ -94,18 +94,19 @@ def notary
...
   # there is more there, including GeoIP which we may use as when we
   # address HT-1451
-  def log_action(log_entry, permitted_params)
+  def log_action(obj, permitted_params)
     raise UnfilteredParameters unless permitted_params.permitted?
 
+    raise "Unable to extract object id from #{obj.inspect}" if obj.id.blank?
     rails_action = "#{params[:controller]}##{params[:action]}"
-
+    log_entry = HTLog.new(model: obj.resource_type.to_s.camelize, objid: obj.id)
```
aelkiss, line 102: I wonder if perhaps we should just pass `obj` to `HTLog.new` and have it be its responsibility to ask the object how it wants to be identified in the logs? e.g something like

```ruby
class HTLog

  def initialize(obj: nil, model: obj.resource_type.to_s.camelize, objid: obj.id)
    ...
  end

end
```

and

```
 Rails.logger.info "AUDIT LOG: #{log_entry.model}, #{log_entry.objid}, #{JSON.generate(log_entry.data)}"
```

app/controllers/ht_contact_types_controller.rb
```diff
@@ -50,6 +54,10 @@ def destroy
 
   private
 
+  def log(params = @contact_type_params)
```
aelkiss, line 57: I started to wonder if there was something clever we could do to define this once at the application controller and infer the correct object & params, but I think doing it the way you are rather than something clever and more Rails-like is  preferable.

aelkiss (commented):
The approach looks good and makes sense. One thought (inline) about passing the object straight to `HTLog` rather than making extraction of type & id the responsibility of `log_action`.

I *think* the current behavior is that someone who can admin anything can view logs, but nobody else can. We aren't testing that though.

We may be rapidly approaching the point where we want to add some kind of policy layer to OTIS.

aelkiss:
Tried it out in dev; works as expected. Eventually we may want some kind of more formatted display, e.g. to view the history of a user etc, but this looks good for now.

It strikes me that for update actions it might be helpful to log the previous value of the object (so we know what specifically changed) but I don't know that we need to do that now; may be worth revisiting at some point, esp. if there's a gem that will do it for us (e.g. https://github.com/collectiveidea/audited)

PR #112, opened 2021-09-01
HT-3129 view and download contacts by type

moseshll:
- Contacts index page can restrict to type via a dropdown menu.
- Add download CSV button.


aelkiss (approved):
Tried it out in development and it looks good; tests look good; let's go ahead and deploy at your convenience.

PR #113, opened 2021-09-10
HT-3158 Remove 'renewed' badge

moseshll:


aelkiss (approved):

aelkiss:
Looks like it does what it says it does. We can see if this is less confusing in the future for doing the renewal workflow.

PR #114, opened 2021-09-27
HT-3159 purge request expired labels

moseshll:
Also change "delete" buttons for contacts and contact types to a "danger" style for consistency with the purge button added to the approval requests index.

There's also a fix for one of the HTInstitution tests that wasn't HTML-escaping the institution name. There are likely others. (Culprit was an unescaped apostrophe.)

### DIFF
--- a/app/controllers/ht_approval_requests_controller.rb
+++ b/app/controllers/ht_approval_requests_controller.rb
@@ -14,6 +14,10 @@ class HTApprovalRequestsController < ApplicationController
       flash[:notice] = "Renewed #{@renewed_users.join ", "}" if @renewed_users.count.positive?
       session[:renewed_users] = @renewed_users
     end
+    if params[:delete_expired]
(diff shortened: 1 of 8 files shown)


aelkiss (commented):
Looks good at first glance. I'll give it a try in development.

Out of scope for this PR (but relevant to the question of institution name escaping) -- we've certainly run into issues with institution names that contain `&`, and it does not look like OTIS is currently handling those as expected - we expect that the institution name is stored with character entities e.g. as `Texas A&amp;M University`, and it looks like that's being double-escaped for display.

moseshll:
Added a late-breaking commit to incorporate a change that was unsaved in my text editor.

aelkiss:
Looks good to me in development. Let's squash the extra commit and then you can go ahead and merge.

aelkiss (approved):

PR #116, opened 2021-10-04
HT-3160 Map from internal terms like 'ssdproxy', 'crms' to readable t…

moseshll:
…erms

- Move descriptions from HTUser model into the locale file.
- Add accessors via HTUserPresenter.


mwarin (approved):
Makes sense to me.

PR #120, opened 2021-10-19
HT-3203, added cols to csv output

mwarin:
Couldn't figure out how to the wanted columns to appear when calling .attributes so I added 2 methods in the ht_user model for making csv header rows and csv value rows.


moseshll (changes requested):
I'd like to see the test also check the institution id and name in the body of the CSV. Just for fun:
```
def setup
    @inst1 = create(:ht_institution, entityID: "http://example.com", inst_id: 'X', name: 'Y')
...
```
then append `",X,Y"` to the string around line 277 to make sure those values are also in the response body.

Other than that it looks fine.

mwarin:
@moseshll I think I addressed your one comment above, so this could be merged.

moseshll (approved):

PR #118, opened 2021-10-06
minimum viable thing

mwarin:
The minimum viable thing contains:
* db table declaration
* model, controller, views
* submittable input form
* seed

PR #121, opened 2021-10-21
User registrations now with simplified database stuff

aelkiss:
Supercedes #118


test/models/ht_registration_test.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class HTRegistrationTest < ActiveSupport::TestCase
+  test "basic math" do
+    assert_equal(1, 1)
+  end
+
+  test "validation passes" do
```
aelkiss, line 10:
```ruby
test "factory builds a valid ht_registration"
```

test/models/ht_registration_test.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class HTRegistrationTest < ActiveSupport::TestCase
+  test "basic math" do
```
aelkiss, line 6: Needs to be removed

aelkiss:
I think this PR is superceded at this point. @mwarin there's a different branch that continues this work (but without the merge conflicts), right?

PR #122, opened 2021-10-26
HT-3054 users with admin ht_users can't admin approval requests

moseshll:
- Add Checkpoint policies derived from controller or resource type (more meta stuff than I would have liked, perhaps).
- Address pluralization mismatch between our grants (`lib/tasks/migrate_users.rake`) and model `#resource_type`.
  -  I don't really like the `#resource_type` pluralization route but this preserves the grants we have in production.
- We still don't do any per-object auth checks (i.e., can I edit myself), but we should now be able to if needed.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -26,9 +26,8 @@ class ApplicationController < ActionController::Base
     raise NotAuthorizedError unless can?(params[:action], params[:controller], current_user)
   end
 
-  def can?(action, resource, user = current_user)
-    res = Checkpoint::Resource::AllOfType.new(resource)
-    Checkpoint::Query::ActionPermitted.new(user, action, res, authority: Services.checkpoint).true?
+  def can?(action, object, user = current_user)
+    policy(object).can?(action, authorizable_object(object), user)
   end
 
   # This could be replaced by a landing page that is accessible
@@ -40,6 +39,17 @@ class ApplicationController < ActionController::Base
   end
 
   private
+
+  # Turn string or symbol referencing ActiveRecord class into that class.
+  # For ActiveRecord objects and classes this is an identity.
+  # Typically this will take something like "ht_users" and return HTUser.
+  # Use this intermediate representation (instead of just converting to a Checkpoint::Resource)
+  # because policies may need to reason about real model objects.
+  def authorizable_object(object)
+    return object if object.respond_to? :to_resource
+
+    Object.const_get(object.to_s.singularize.camelize)
+  end
 
   def user_not_authorized
     # logout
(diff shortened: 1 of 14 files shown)


moseshll:
@aelkiss My main objection to using a custom resource resolver is that it requires injecting a to-Checkpoint conversion into the `HTApprovalRequest` model in order to be used globally. The docs seem to suggest using a resolver or policies, not necessarily both.

I'm leaning in the direction of making `can?` require a generic `Resource` object that can keep most of the `is_a?` garbage in one place, preferably outside the policy code.

aelkiss:
I'll try to look at this again tomorrow -- sorry for the delay.

moseshll:
A simple Resource class that can hide some of the `is_a?` nonsense.

https://gist.github.com/moseshll/21bfaa9f88306541fc25574350e5eb9b

botimer:
Hmmm... I'm dubious of a base class or wrappers. The design intent throughout has been that you can use your domain/persistent objects as the resources -- with a minimal interface to implement for adaptation. I think it's good that you are wary of duplicating this conditional behavior, but it's probably indicating a different intervention.

I have perhaps not been tracking this as closely as I should. I think you're addressing two issues: that you want to work with definite and indefinite resources, and that you want some convenience for referring to the indefinite ones. In technical terms, that's roughly: you want to work with both resource instances and classes, and you want a handy reference to the class you can use without an instance -- all the while not polluting the domain with too much framework stuff.

Let's discuss in real time; maybe something will emerge that satisfies both these needs and my possibly ill-informed design sensibilities on the issue.

moseshll:
I did apparently miss one opportunity to use `to_resource_name` in the `Otis::Authorization::Resource` module. I suspect I got distracted halfway through the method extraction. Irritating but not a show-stopper.
This is re-based on the registrations code. I have not given that much thought to business rules on registrations. It prolly belongs in another ticket.

aelkiss (approved):
I'm pretty happy with this. There are only a few concrete places we bind in the code (rather than tests) to Checkpoint, the policies don't have to worry about individual objects vs. collections of objects, and the approval requests policy is expressed in terms of the business logic rather than Checkpoint logic.

We can try the magic policy-finding for now. I suspect we may find it more manageable later on down the road to do something like

```ruby
class ApplicationController
  def policy
    @policy ||= ApplicationPolicy
  end
  
  attr_writer :policy
end

class RegistrationController < ApplicationController
  def policy
    @policy ||= RegistrationPolicy
  end
end
```

That pattern also allows us to more easily inject a policy in tests (I'm struggling to remember exactly why I wanted to do that in chipmunk, but I remember it was a big frustration with controller testing)

botimer:
Yup, I agree. And as far as convention vs configuration, a decade of Rails programming has taught me: I don't mind a couple of lines of configuration for the clarity that the breadcrumbs give you.

aelkiss:
@moseshll  Ready to merge?

PR #128, opened 2021-11-12
Ht 3155 registrations

mwarin:


mwarin:
Tests exist & pass, commits squorshed.

moseshll (approved):
Test coverage seems good. I like the fetch_* controller methods and will probably steal the idea.

aelkiss (approved):
We should make sure we understand how the sign-up process varies for non-ATRS users, but this looks like it covers the ticket as described.

It may be worth checking whether the current user can create a registration when deciding whether to show the 'new registration' button or not -- not a big deal for now though.

moseshll:
I'm going to treat auth as out of scope for this ticket simply because so much is still up in the air in that department. With the understanding that there is likely some fix-up needed on this I can do under HT-3054. I'll plan to merge this after my call with Aaron.

PR #129, opened 2021-12-01
HT-3224: Rename database tables

aelkiss:
This doesn't rename any of the classes in the application, just the
database tables themselves.

Eventually, I think it would make sense to put everything under an `Otis` namespace, e.g.

```ruby
module Otis
  class Registration
   ...
  end
end
```

That requires bigger changes and will almost certainly cause merge conflicts with any other work in progress, so we should probably discuss and coordinate that a bit more when/if we want to do that.

### DIFF
--- a/app/models/ht_approval_request.rb
+++ b/app/models/ht_approval_request.rb
@@ -2,6 +2,8 @@
 
 class HTApprovalRequest < ApplicationRecord
   self.primary_key = "id"
+  self.table_name = "otis_approval_requests"
+
   def self.expiration_date
     Date.today - 1.week
   end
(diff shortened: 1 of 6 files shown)


aelkiss:
I've already renamed the tables in production and created both `otis_` and `ht_` views, so this should be safe to deploy any time.

moseshll (approved):
I think you're spot on re using an OTIS namespace eventually.
This all looks fine.

aelkiss:
I'll go ahead and merge; that should automatically deploy to staging; if that all works I'll deploy to production on Monday.

PR #130, opened 2021-12-06
Singularize resources for grants

aelkiss:
As far as I can tell, we should now singularize the resource types for
grants, and we can use the class names when calling 'can?'.

There are a number of places wehere we could clean things up by removing
some logic from views and potentially simplifying that logic through
policies, but but changing that is out of the scope of this commit.


moseshll (approved):
I like dropping the distracting indirection between grants and classes. Looks good here.

PR #132, opened 2022-01-07
HT-1495 Migrate OTIS UI strings from views to config/locales/en.yml

moseshll:
- First off, English UI strings should be about 99% migrated out of MVC. For testing I have produced a Japanese locale file (`config/locale/ja.yml`) that should only be used for development/testing. It is not enabled in `config/initializers/locale.rb` because it is only slightly better than the documentation you get with cheap imported electronics. I chose it because having a non-Roman locale makes it easier to spot missing translations. If having bad pidgin Japanese in the repo is a potential embarrassment to HathiTrust, I suggest replacing it with a pseudo-locale based on "upside down" characters -- that could be auto-generated.
- The approval request mailer is not localized, so some UI strings that could be excised from e.g. the `ExpirationDate` class have been left intact. Because the preferred language of the OTIS user is orthogonal to the preferred language of the mail recipient, I conclude that there needs to be (if mailers are to send anything non-English) some kind of an additional switch for the mailer. Pushing and popping I18n.locale doesn't seem to be hard (I think I do it in one of the tests) so I don't think Rails would fight us too hard if we wanted to custom-localize a partial.
- Japanese provides a good argument against using checkmark icons to indicate Boolean `true`: in Japan a checkmark or an X means a test answer is incorrect. I have gone a little badge-crazy in response as badges provide both color and natural language clues. (And yes, I know we shouldn't be using green vs red.) Note the `HTInstitution` "US" field now uses a flag icon.
- Attacking the accumulated proliferation of ad-hoc presenters made it clear that we have too many different names for things. And it's hard to keep fields in consistent order between show pages and editing forms. `ApplicationPresenter` makes some attempt to fix both of those. It is mainly concerned with getting model data into a page in an appropriate form (localized, in other words). It is less concerned with what buttons look like. Presenters are too tightly coupled with controllers, but they're a necessary evil since Rails' I18n routines don't work properly without a controller somewhere in the mix.
- Test coverage is now hovering just under 98%. I sped up the normally-skipped HTML validation stuff by shuffling the page load off into a block and bailing out early if validation is disabled. Tests still assume the UI language is English. In the end I didn't add much to the tests in terms of locale awareness. That's a whole 'nother thing if we decide to go the locale testing route.
- The logic around displaying check boxes in the Users index page table is still in a public `HTUserPresenter` method and it's a mess. I may attack that a bit more before finalizing the PR. (EDIT: I didn't because I'd like to move on to something else for a while. But at some point it would be worth doing.)
- I've gone the route of using exceptions rather than nathural language strings to signal data validation errors. `IPRestrictionError` is a good example of this, since we can stick the appropriate localization key inside the error.

~~This branch still needs to be rebased off changes from HT-3261.~~


app/views/ht_registrations/show.html.erb
```diff
@@ -1,39 +1,20 @@
...
-    <dd><%= @registration.dsp_name %></dd>
-    <dt>DSP Email:</dt>
-    <dd>
-      <%= link_to(@registration.dsp_email, "mailto:#{@registration.dsp_email}") %>
-    </dd>
-    <dt>DSP Date:</dt>
-    <dd><%= @registration.dsp_date %></dd>
-    <dt>MFA Addendum:</dt>
-    <dd><%= @registration.mfa_addendum %></dd>
+    <% HTRegistrationPresenter::ALL_FIELDS.each do |field| %>
```
aelkiss, line 5: My inclination would be to get `ALL_FIELDS` from the instance rather than binding to the specific class name here.

app/presenters/application_presenter.rb
```diff
@@ -0,0 +1,115 @@
...
+    method = ("show_" + field.to_s).to_sym
+    if respond_to?(method, true)
+      send(method) || ""
+    else
+      localize_value field
+    end
+  end
+
+  def localize_value(field)
+    value = try(field) || return
```
aelkiss, line 95: This will call the method `field` in the object if it exists; we're not sure what the use case there is.

app/presenters/application_presenter.rb
```diff
@@ -0,0 +1,115 @@
...
+
+  def initialize(obj, controller: nil, action: :show)
+    super obj
+    @controller = controller
+    @action = action
+  end
+
+  # By default form cancel button goes to show page if persisted,
+  # index page otherwise.
+  def cancel_path
```
aelkiss, line 45: Not sure the metaprogramming/class name manipulation is worth the DRY savings here.

app/controllers/application_controller.rb
```diff
@@ -38,6 +60,10 @@ def default_path
     root_path
   end
 
+  def default_url_options
```
aelkiss, line 63: This ends up sticking `?locale=en` etc on the end of every link the application -- is that recommended as best practice? If so I'll go with it, but my inclination would be to instead set it on the session or something instead.

app/presenters/application_presenter.rb
```diff
@@ -0,0 +1,115 @@
...
+    @value_scope ||= name.sub(%r{Presenter$}, "").underscore + ".values"
+  end
+
+  def self.field_label(field)
+    I18n.t(field, scope: attribute_scope)
+  end
+
+  def initialize(obj, controller: nil, action: :show)
+    super obj
+    @controller = controller
```
aelkiss, line 39: Is the reason we need the controller here so that we have the URL helpers available?

aelkiss (approved):
I took another look through and I'm reasonably happy that I understand how things work. The abstractions in the presenters make sense to me and I think the responsibility is appropriate; I don't love that they need to know about the controllers but (assuming the reason is that so they can use the URL helpers) that seems difficult to avoid and more of a product of the way Rails controllers work than a consequence of our design.

app/controllers/application_controller.rb
moseshll: It's the behavior I noticed when dealing with Fulcrum/Heliotrope. I don't know if that's from Samvera or something the team did locally.

moseshll:
It will probably be easier if I manually rebase the conflicts locally and then do another force push. I'm bumping this up to my top priority since this is starting to be a blocker.

app/presenters/application_presenter.rb
aelkiss: :+1:

app/presenters/application_presenter.rb
moseshll: I think I can explain the reason I used `try(f)` instead of `send(f)`: for derived or aggregate fields that are not model methods (for example the composite `auth_rep` in `HTRegistration` that is composed of the three model name/email/date fields) when the presenter's `show_X` method hasn't been written yet. (The `show_X` bleeds off the `localize_value` logic.) So `send(:auth_rep)` would throw an exception where `try(:auth_rep)` would return `nil`, and `localize_value` would just return `nil` at that point. So the value of `try` is to get something running in a development scenario where the presenter may not be fleshed out. I'll leave it up to @aelkiss whether to comment and keep,  or replace with `send`.

moseshll: TL;DR
If my presenter lists (in e.g., `ALL_FIELDS`) a field `:blah` and we're rendering it.
1. Call `self.show_blah` if it's implemented (implementation is responsible for getting the value).
2. Otherwise call `self.try :blah` or `self.send :blah` (the latter potentially throwing an exception) to get the value.
3. Check for a localized version of the value with `t(self.class.value_scope + ".#{field}.#{value}"`
4. HTML escape the raw value if all else fails (which is perfectly legit).

app/presenters/application_presenter.rb
moseshll: It's nightmarish to read, but if it saves someone typing mind-numbing boilerplate...

app/presenters/application_presenter.rb
moseshll: Removed in forthcoming commit.

app/controllers/application_controller.rb
moseshll: Re path component approach: just dropping in the routing changes suggested by the Rails I18n guide appears to break URL generation when we use the `some_path some_model_object` convention for generating URLs. There may also be interference from `RAILS_RELATIVE_URL_ROOT`. I have a WIP to experiment with, but that's for another ticket.

moseshll:
@aelkiss Will merge 02/11 unless there are outstanding issues.

app/presenters/application_presenter.rb
aelkiss: I think fine to keep for now -- if we run into strange issues debugging stuff or get ourselves confused in the future we might find that we would rather get rid of this and just have the boilerplate, e.g.:

```ruby
class ApplicationPresenter
  def cancel_path
    if persisted?
      send show_route_helper, self
    else
      send index_route_helper
    end
   end

end

class WhateverPresenter < ApplicationPresenter
  def show_route_helper
    :whatever_path
    # or maybe :whatever_path.to_proc
  end

  def index_route_helper
    :whatevers_path
  end
end
```

There might be some way to use `url_for` to get those urls out. I have often wished there was something defined on the model that gave you the name of the route helpers, but I've never been able to find it.

In any case, I'm fine with this as-is for now, and hopefully we won't have to touch it..?

aelkiss:
I'm fine with merging this as-is unless you particularly want to try to do anything different with `cancel_path`.

app/presenters/application_presenter.rb
moseshll: I've been navigating back and forth checking the registrations in particular, sending mails and checking the landing pages, no issues there as far as using the cancel button for navigation. Hoping we don't have to touch it for a while. I'm gonna leave it in place.

PR #133, opened 2022-01-14
HT-3304 ht_users should join on inst_id, not entityID

moseshll:
- Add a new column `ht_users.inst_id` while deprecating `ht_users.identity_provider`.
- Do not enforce validity of `inst_id` when creating users since it breaks a number of tests. Right now the UI doesn't allow editing, and registrations restrict selection to menu of institutions, so I'm not too concerned about that right now.
- **This will require a production schema change before rollout.**
  - CRMS uses the field in a routine that appears not to be called currently but this needs to be investigated a bit more deeply.
  - Check in with Roger and Bill for possible issues with MDP lib/tools or whatever might lurk.

### DIFF
--- a/test/controllers/ht_users_controller_test.rb
+++ b/test/controllers/ht_users_controller_test.rb
@@ -254,13 +254,13 @@ end
 
 class HTUsersControllerCSVTest < ActionDispatch::IntegrationTest
   def setup
-    @inst1 = create(:ht_institution, entityID: "http://example.com", inst_id: "X", name: "Y")
+    @inst1 = create(:ht_institution, inst_id: "X", name: "Y")
     @user1 = HTUser.new(userid: "a@b", displayname: "A B", email: "c@d",
       activitycontact: "e@f", approver: "g@h",
       authorizer: "i@j", usertype: "staff", role: "ssd",
       access: "total", expires: "2020-01-01 00:00:00",
       expire_type: "expiresannually", iprestrict: "any",
-      mfa: false, identity_provider: "http://example.com")
+      mfa: false, identity_provider: "http://example.com", inst_id: "X")
     @user1.save!
     @user2 = create(:ht_user, :expired, userid: "y@z")
   end
--- a/app/models/ht_institution.rb
+++ b/app/models/ht_institution.rb
@@ -5,7 +5,7 @@ class HTInstitution < ApplicationRecord
   accepts_nested_attributes_for :ht_billing_member, update_only: true
 
   self.primary_key = "inst_id"
-  has_many :ht_users, foreign_key: :identity_provider, primary_key: :entityID
+  has_many :ht_users, foreign_key: :inst_id, primary_key: :inst_id
   has_many :ht_contacts, foreign_key: :inst_id, primary_key: :inst_id
   has_many :ht_registrations, foreign_key: :inst_id, primary_key: :inst_id
   has_many :ht_logs, -> { HTLog.ht_institution }, foreign_key: :objid, primary_key: :inst_id
--- a/app/models/ht_user.rb
+++ b/app/models/ht_user.rb
@@ -6,7 +6,7 @@ require "forwardable"
 class HTUser < ApplicationRecord
   self.primary_key = "email"
 
-  belongs_to :ht_institution, foreign_key: :identity_provider, primary_key: :entityID
+  belongs_to :ht_institution, foreign_key: :inst_id, primary_key: :inst_id
   has_one :ht_count, foreign_key: :userid, primary_key: :userid
   has_many :ht_logs, -> { HTLog.ht_user }, foreign_key: :objid, primary_key: :email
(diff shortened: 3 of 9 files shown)


aelkiss:
I'll aim to look at this Tuesday. In the meantime @billdueber  @respinos  Are either of you aware of any places that use `ht_users.identity_provider` ?

billdueber:
I can't find any uses in catalog, ls, or pt.  [Searching the organization](https://github.com/search?q=org%3Ahathitrust+identity_provider&type=code) gives a few hits. Most of these, I think, are getting the data from the Shib session. 

One that definitely uses that column is a direct SQL call in  [mdp-lib ACL logic](https://github.com/hathitrust/mdp-lib/blob/0053cb982db7b91215eb7b6a5ff9daee8c202c5e/Auth/ACL.pm#L592-L604). It was added by @respinos five years ago and is used if `$do_restrict_to_identity_provider` is true.

test/controllers/ht_users_controller_test.rb
```diff
@@ -254,13 +254,14 @@ def setup
...
-    @inst1 = create(:ht_institution, entityID: "http://example.com", inst_id: "X", name: "Y")
+    @inst1 = create(:ht_institution, entityID: "http://counterexample.com",
+      inst_id: "X", name: "Y")
     @user1 = HTUser.new(userid: "a@b", displayname: "A B", email: "c@d",
       activitycontact: "e@f", approver: "g@h",
       authorizer: "i@j", usertype: "staff", role: "ssd",
       access: "total", expires: "2020-01-01 00:00:00",
       expire_type: "expiresannually", iprestrict: "any",
-      mfa: false, identity_provider: "http://example.com")
+      mfa: false, identity_provider: "http://example.com", inst_id: "X")
```
aelkiss, line 264: Not sure I understand 1) why the user still lists an `identity_provider` and 2) why this test passes...

respinos: > I can't find any uses in catalog, ls, or pt. [Searching the organization](https://github.com/search?q=org%3Ahathitrust+identity_provider&type=code) gives a few hits. Most of these, I think, are getting the data from the Shib session.
> 
> One that definitely uses that column is a direct SQL call in [mdp-lib ACL logic](https://github.com/hathitrust/mdp-lib/blob/0053cb982db7b91215eb7b6a5ff9daee8c202c5e/Auth/ACL.pm#L592-L604). It was added by @respinos five years ago and is used if `$do_restrict_to_identity_provider` is true.

`$do_restrict_to_identity_provider` is an obsolete mechanism of ensuring a friend account couldn't appear as one of the institutions that was using email addresses as the identifier. (We decided not to do this.)

respinos:
> I can't find any uses in catalog, ls, or pt. [Searching the organization](https://github.com/search?q=org%3Ahathitrust+identity_provider&type=code) gives a few hits. Most of these, I think, are getting the data from the Shib session.
> 
> One that definitely uses that column is a direct SQL call in [mdp-lib ACL logic](https://github.com/hathitrust/mdp-lib/blob/0053cb982db7b91215eb7b6a5ff9daee8c202c5e/Auth/ACL.pm#L592-L604). It was added by @respinos five years ago and is used if `$do_restrict_to_identity_provider` is true.

`$do_restrict_to_identity_provider` is an obsolete mechanism of ensuring a friend account couldn't appear as one of the institutions that was using email addresses as the identifier. (We decided not to do this.)

test/controllers/ht_users_controller_test.rb
moseshll: It's a roundabout way of testing the join, with `example.com` changed to `counterexample.com`. This test would fail if the code were still insisting on a join via identity provider. This is one of the tests that initially failed with the new join, and for whatever reason I didn't bother to create a more explicit test after tweaking it.

moseshll:
I have a PR in for CRMS to remove access to `identity_provider` since it was not being used.
@respinos sounds like nothing really needs to take place in mdp-lib since `$do_restrict_to_identity_provider` is always `0`. Unless you think it is time to strip out that code entirely?

respinos:
> I have a PR in for CRMS to remove access to `identity_provider` since it was not being used. @respinos sounds like nothing really needs to take place in mdp-lib since `$do_restrict_to_identity_provider` is always `0`. Unless you think it is time to strip out that code entirely?

We should probably schedule the mdp-lib update independently.

moseshll:
CRMS and db-image have been updated.

test/controllers/ht_users_controller_test.rb
moseshll: I added an explicit test in `models/ht_users_test.rb` and dialed back the unnecessary stuff here by removing the institution's `entityID` as irrelevant. That should make this test a little less head-scratchy.

moseshll:
I think this should be ready to go modulo database tweaking.

aelkiss:
Database schema updated in production; verified working in testing. Will release to production.

PR #134, opened 2022-01-19
HT-3154 As a HathiTrust staff member, I want to send an access key to…

moseshll:
… a user who has filled out a form for elevated access, so that they can get access.

- Implementation based on approval requests mailer.
- Some re-jiggering of otis_registrations fields based on ticket comments, e.g.: contact info no longer refers to the user being registered.
- This is still somewhat ATRS specific.

### DIFF
--- a/app/views/registration_mailer/registration_email.text.erb
+++ b/app/views/registration_mailer/registration_email.text.erb
@@ -1,0 +1,1 @@
+<%= strip_tags @body %>
--- a/README.md
+++ b/README.md
@@ -37,3 +37,9 @@ These tests are not run by default since they rely on an external service.
 ```
 docker-compose run -e W3C_VALIDATION=1 test
 ```
+
+To run a single test class use an invocation along these lines:
+
+```
+docker-compose run test bundle exec ruby -I test test/controllers/ht_users_controller_test.rb
+```
--- a/app/controllers/ht_registrations_controller.rb
+++ b/app/controllers/ht_registrations_controller.rb
@@ -31,8 +31,8 @@ class HTRegistrationsController < ApplicationController
     @registration = HTRegistrationPresenter.new(HTRegistration.new(reg_params))
     if @registration.save
       log
-      flash.now[:alert] = "Registration created for #{@registration.name}."
-      redirect_to action: :index
+      flash.now[:alert] = "Registration created for #{@registration.dsp_name}."
+      redirect_to preview_ht_registration_path(@registration)
     else
       flash.now[:alert] = @registration.errors.full_messages.to_sentence
       render "new"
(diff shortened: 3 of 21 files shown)


moseshll:
Noting here that this branch will set off merge conflicts with the HT-1495 I18n branch so there will be a round of manual resolution. I didn't try to make this branch I18n aware.

app/views/registration_mailer/registration_email.text.erb
```diff
@@ -0,0 +1 @@
+<%= strip_tags @body %>
```
aelkiss, line 1: I think using `strip_tags` here means that the link with the token won't be visible in the output email, so the text version won't be usable. It might be worthwhile including only the common text in the body and include differently-formatted links in the HTML and text version?

aelkiss:
Right now the workflow matches renewals in terms of first having to create the renewal or registration, then select which one, then email. I think that works OK for renewals because generally we're renewing multiple users, then emailing each approver. For registrations, it might make more sense to take the user to the email preview right away after creating the registration rather than going back to the index, or at least to go to the show page (where it's just one more click to go to the e-mail preview)

app/views/registration_mailer/registration_email.text.erb
moseshll: Yes indeed. At the very least those links need to be readable as opposed to `<a ...>click here</a>` formatting.

I think I'd rather see the whole e-mail previewed and editable in the preview page, otherwise there's hidden stuff going into the outgoing mails that is either invisible to the editor or displayed around the ckeditor pane.

Let's see what I can do with the source HTML template. I haven't dealt with previewing Mailer output since we first worked on the approval requests, but I need to add to `test/mailers/previews` for starters. Then I can get a better idea about what needs to be done.

moseshll:
Agree with the desire for new -> show -> preview workflow.

app/views/registration_mailer/registration_email.text.erb
aelkiss: Good point regarding seeing the entire email in the preview.

moseshll: I've tried to push the parameters inside the mailer towards more "leaf node" than primitive (specify finalize URL rather than deriving it from the registration, the controller, and God knows what else inside the mailer). This approach seems more preview-friendly since Rails' Action Mailer preview functionality seems to be a duct-taped collection of hacks.

moseshll: The text preview appears to be acceptably equivalent to the HTML.
http://localhost:3000/useradmin/rails/mailers/registration_mailer/registration_email
... since I can never remember the URL

moseshll:
Force-pushed a new branch incorporating a fix for the finalize URL displayed in the in-app mail preview (not the Rails test preview I referenced and griped about previously).

aelkiss (approved):
OK, I took another look & tried it in development; looks good to me now.

PR #135, opened 2022-02-08
HT-3156 landing page to confirm user registrations

moseshll:
- Further patterned after what we do with approval requests.
- Not sure about the wording for different MFA/non-MFA scenarios. Expect some cleanup with HT-1212.
- Some new logic for SEND/RESEND link in the e-mail preview.

### DIFF
--- a/app/views/shared/finalize.html.erb
+++ b/app/views/shared/finalize.html.erb
@@ -1,0 +1,43 @@
+<div id="maincontent" class="row">
+
+  <% if @registration.expired? %>
+    <strong>Registration for <%= @registration.dsp_email %> has expired.
+            Please contact <%= mail_to ApplicationMailer.default[:from] %>.</strong>
+  <% elsif @already_used %>
+    <strong>This link is no longer valid; it may have already been used.</strong>
+  <% else %>
+    <p><i>WARNING: THE FOLLOWING PLACEHOLDER TEXT HAS NOT BEEN REVIEWED BY THE LEGAL DEPARTMENT.</i></p>
+    <p>Thank you.</p>
+    <p>Elevated access registration confirmed for <strong><%= @registration.dsp_email %></strong>.</p>
+    <p>A HathiTrust staff member will approve your registration and respond shortly by e-mail.</p>
+    
+    <% if @institution.entityID && @institution.shib_authncontext_class %>
+
+      <p>Your registration indicates you have access to multi-factor authentication (MFA).
+        <i>As a reminder, this means than any secure access to HathiTrust must be done
+           through you institution's MFA-enabled sign-in portal.</i>
+      </p>
+      <p>You can test your ability to log in with MFA using the link
+        <%= raw @institution.mfa_test_link %>.
+      </p>
+
+    <% elsif @registration.mfa_addendum? %>
+
+      <% # TODO HT-1212: we want to direct users out and back to pick up MFA info if possible %>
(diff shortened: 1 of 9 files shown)


app/views/shared/finalize.html.erb
```diff
@@ -0,0 +1,30 @@
...
+    <p>Elevated access registration confirmed for <strong><%= @registration.dsp_email %></strong>.</p>
+    <p>A HathiTrust staff member will approve your registration and respond shortly by e-mail.</p>
+    
+    <% if @institution.entityID && @institution.shib_authncontext_class %>
+      <p>Your registration indicates you have access to mult-factor authentication (MFA).</p>
+      <p>You can test your ability to log in with multi-factor authentication using the link
+        <%= raw @institution.mfa_test_link %>.
+    <% elsif @registration.mfa_addendum? %>
+      <% # TODO HT-1212: we want to direct users out and back to pick up MFA info if possible %>
+      <p>Your registration indicates you have access to mult-factor authentication (MFA); however,
```
aelkiss, line 19: The MFA addendum means the user attested they used MFA to log in to HathiTrust, but that we can't detect that via the auth context class. I don't think we need the language about 'please contact' but some reminder from the MFA addendum form about the expectations for MFA might be helpful.

app/views/shared/finalize.html.erb
```diff
@@ -0,0 +1,30 @@
...
+         we do not currently have your institution configured for MFA access.
+         Please contact <%= mail_to ApplicationMailer.default[:from] %>.</p>
+    <% else %>
+      <p>Please take a moment to confirm that <code><%= raw @ip_address %></code> is a static IP address.
```
aelkiss, line 23: Might need to consider adding: This is the only IP address from which you will be able to use [the service, whatever we call it.] If your IP address changes in the future, you will need to [follow some process to change it]

Fine for this to be placeholder text that we review later.

app/views/shared/finalize.html.erb
```diff
@@ -0,0 +1,30 @@
...
+         If you are not sure, we recommend you contact your network administrator.</p>
+      <p>You can test your ability to log in at HathiTrust using the link
+        <%= @institution.login_test_link %>.</p>
+      <pre><%= @whois.to_s %></pre>
```
aelkiss, line 27: I had thought we were going to show the whois on the page where HT staff approved their registration -- was there a reason to show it here also? I would worry that it would be a little bit overwhelming for users since we can't easily parse it just to organization.

aelkiss (approved):
Logic and tests look good for now; there could be some future opportunity with refactoring for `ApprovalController` and `FinalizeController` but they are short enough I'm not concerned about the duplication for now, and I think they're just as likely to diverge -- we can perhaps consider revisit when we have the full workflow in place.

The text on the finalization screen almost certainly needs some work, and it's worth noting that in the ticket for reviewing the email text, etc. I don't think that needs to block merging the PR, though.

app/views/shared/finalize.html.erb
moseshll: Quoting the ticket, which I dutifully copy-pasted into the source file:  "show the IP address (and perhaps the WHOIS info?)" in reference to the submission page for the prospective user. But yeah it's pretty overwhelming. I haven't gotten as far as envisioning the UI for our staff to review the "signed agreements" which is where indeed some kind of whois info is called for (in a tab maybe?).

PR #138, opened 2022-02-16
HT-1194 HathiTrust staff member can finalize registration in ht_users

moseshll:
- Add app/lib/otis/registration_mover.rb for stuffing registration data into new user.
- Add ability to edit ht_user usertype, role, access for newly-registered users.
- Display WHOIS data on registration show page for received registrations (we have the IP address).

### DIFF
--- a/app/controllers/finalize_controller.rb
+++ b/app/controllers/finalize_controller.rb
@@ -11,10 +11,7 @@ class FinalizeController < ApplicationController
     return render_not_found unless @registration&.token_hash
 
     @already_used = @registration.received.present?
-    @ip_address = request.remote_ip
-    # WHOIS text is not appropriate content for a signup page.
-    # This can be moved into a <pre> block in a ht_registration -> ht_user transmogrification page.
-    # @whois = Whois::Client.new.lookup(@ip_address)
+    @ip_address = ip_address
     @institution = HTInstitutionPresenter.new @registration.ht_institution
     @message_type = if @institution.entityID && @institution.shib_authncontext_class
       :success_mfa
@@ -35,9 +32,22 @@ class FinalizeController < ApplicationController
 
   def finalize
     @registration.received = Time.zone.now
+    @registration.ip_address = @ip_address
+    @registration.env = extract_shib_env
     @registration.save!
     # Currently, there are no parameters for the controller other than the
     # token, which we do not wish to log.
     log_action(@registration, params.permit)
+  end
+
+  # For development and (maybe) testing, visiting this page will taint
+  # the ip_address field with "localhost" and cause a validation error on ht_user
+  # created with this registration. So we use a TEST-NET-1 value.
+  def ip_address
+    Rails.env.production? ? request.remote_ip : "192.0.2.1"
+  end
+
+  def extract_shib_env
+    request.env.select { |k, _v| k.match(/^HTTP_X_SHIB/) || k == "HTTP_X_REMOTE_USER" }.to_json
   end
 end
--- a/app/controllers/ht_registrations_controller.rb
+++ b/app/controllers/ht_registrations_controller.rb
@@ -39,6 +39,9 @@ class HTRegistrationsController < ApplicationController
 
   def show
     fetch_presenter
+    if @registration.ip_address.present?
+      @whois = Services.whois.lookup(@registration.ip_address)
+    end
   end
 
   def edit
@@ -50,7 +53,7 @@ class HTRegistrationsController < ApplicationController
     if @registration.update(reg_params)
       log
       flash[:notice] = t(".success", name: @registration.dsp_name)
-      redirect_to action: :index
+      redirect_to action: :show
     else
       flash.now[:alert] = @registration.errors.full_messages.to_sentence
       render "edit"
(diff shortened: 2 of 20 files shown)


aelkiss:
Are we capturing the users' SAML session attributes somewhere when they confirm? If not, we should, and we should display those on the confirmation page as well -- if nothing else we need to know the REMOTE_USER and the institution; I think it would be helpful in terms of confirming identity to capture the other SAML attributes as well.

aelkiss:
I think that can be an add-on issue though and doesn't need to block merging this.

I took a look in dev & the workflow looks good. I'll go through the code & tests soon.

app/lib/otis/registration_mover.rb
```diff
@@ -0,0 +1,29 @@
...
+  # an HTRegistration, and saving.
+  class RegistrationMover
+    def initialize(registration)
+      @registration = registration
+    end
+
+    def ht_user
+      return @ht_user unless @user.nil?
+
+      @ht_user = HTUser.new(userid: @registration.dsp_email,
```
aelkiss, line 12: The userid won't always be `@registration.dsp_email`; instead, we need to capture that from the `X-Remote-User` header from finalization, and we should capture other headers for display (at least).

moseshll: Let me make sure I get this right, REMOTE_USER becomes ht_user.userid, for the email we can still use otis_registration.dsp_email or even HTTP_X_SHIB_EDUPERSONPRINCIPALNAME. I was thinking I can either store the HTTP request ENV in a JSON structure in a new field, let's call it otis_registration.finalize_env. Are we storing/displaying anything and everything of the form HTTP_X_SHIB_* in the headers? Anything else (beside REMOTE_USER)?

aelkiss: `HTTP_X_SHIB_EDUPERSONPRINCIPALNAME` won't reliably be populated or (if it is) reliably an email address or reliably the correct email address, so I would still use `otis_registration.dsp_email`. I don't think we need to store anything from the environment besides `HTTP_X_SHIB_*` and `REMOTE_USER`.

aelkiss: It might (?) be preferable to just store the keycard user attributes, as that already takes care of getting them out of the headers and decoding as necessary. OTOH we were thinking about getting rid of keycard entirely so :shrug:

test/controllers/ht_registrations_controller_test.rb
```diff
@@ -72,6 +72,22 @@ def setup
     assert_match(edit_btn, @response.body)
     assert_match(del_btn, @response.body)
   end
+
+  test "received registration shows WHOIS info" do
```
aelkiss, line 76: This hits an external service, so we might want to limit when we run this, similarly to the HTML validation

app/controllers/finalize_controller.rb
```diff
@@ -35,9 +32,17 @@ def authorize!
...
     # Currently, there are no parameters for the controller other than the
     # token, which we do not wish to log.
     log_action(@registration, params.permit)
   end
+
+  # For development and (maybe) testing, visiting this page will taint
+  # the ip_address field with "localhost" and cause a validation error on ht_user
+  # created with this registration. So we use a bogus value.
+  def ip_address
+    Rails.env.production? ? request.remote_ip : "8.8.8.8"
```
aelkiss, line 46: We might prefer to use something in the `TEST-NET-1` block (reserved for documentation & examples) here, e.g. 192.0.2.1 (https://en.wikipedia.org/wiki/Reserved_IP_addresses) rather than a real (albeit well-known) IP address.

test/controllers/ht_registrations_controller_test.rb
```diff
@@ -299,3 +315,39 @@ def mail_registration(registration: @registration, params: {})
     assert_equal "show", @controller.action_name
   end
 end
+
+class HTRegistrationFinishTest < ActionDispatch::IntegrationTest
+  def setup
+    @received_registration = create(:ht_registration, received: Time.now,
+      ip_address: "8.8.8.8")
```
aelkiss, line 322: `Faker::Internet.public_ip_v4_address`, maybe?

app/controllers/ht_registrations_controller.rb
```diff
@@ -39,6 +39,9 @@ def create
 
   def show
     fetch_presenter
+    if @registration.ip_address.present?
+      @whois = Whois::Client.new.lookup(@registration.ip_address)
```
aelkiss, line 43: We might want to avoid doing this in testing; I would recommend in `config/intializers/services.rb`:

```ruby
Services.register(:whois) { Whois::Client.new }
```
and here
```ruby
@whois = Services.whois.lookup(@registration.ip_address)
```
and in `test_helper.rb` (just at the top level, I think; there's no equivalent in minitest that I could find of rspec's `before(:all)`):

```ruby
Services.register(whois) do
  Class.new do
    def lookup(ip)
      "TOTALLY LEGIT WHOIS for #{ip}"
    end
  end.new
end
```
(This creates an instance of a new anonymous class that serves as a test double for the real whois service)

Alternatively, you could look into minitest's mocking/stubbing support in setup for registrations controller tests, but I think I prefer this more explicit dependency injection approach.

aelkiss (changes requested):
See comments inline. I think avoiding hitting the real whois servers in the tests is probably prudent. We definitely at least need to capture the `REMOTE_USER` so we can appropriately set it in the `ht_user` record, and if we're doing that perhaps we might as well capture the other SAML session info for display on finalization as well?

app/lib/otis/registration_mover.rb
aelkiss: :+1:

aelkiss (approved):
Looks good; using the links in dev is a little bit fussy, but I was able to test by:

- using firefox dev tools
- copying the 'request as curl' to get the cookies, etc
- changing the url to the one displayed after sending the email
- adding an `X-Remote-User` and `X-Shib-Whatever` header
- verifying these show up when viewing the registration

Definitely something it would be easier to test end-to-end with the full shib stack. Either way, we'll want to make sure the behavior is as we expect in testing before we deploy this for real.

PR #140, opened 2022-03-07
HT-3360 Keyboard-aware institution pop-up menu

moseshll:
- Use Select2 (select2.org) which appears to be the JS used by HT WAYF.
- Also use Select2 locale files for each of our available locales.
- This facility is currently used only for institution select menus used with registrations and contacts.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -9,6 +9,13 @@
     <%= stylesheet_link_tag    'application', media: 'all', 'data-turbolinks-track': 'reload' %>
     <%= javascript_include_tag 'application', 'data-turbolinks-track': 'reload' %>
 
+    <% # Select2 (select2.org) main distro, locale files, CSS %>
+    <%= javascript_include_tag "https://cdn.jsdelivr.net/npm/select2@4.1.0-rc.0/dist/js/select2.min.js" %>
+    <% I18n.available_locales.each do |locale| %>
+      <%= javascript_include_tag "https://cdn.jsdelivr.net/npm/select2@4.1.0-rc.0/dist/js/i18n/#{locale.to_s}.js" %>
+    <% end %>
+    <%= stylesheet_link_tag "https://cdn.jsdelivr.net/npm/select2@4.1.0-rc.0/dist/css/select2.min.css" %>
+
     <%= stylesheet_link_tag("https://unpkg.com/bootstrap-table@1.18.3/dist/bootstrap-table.min.css") %>
 
     <% # Keyboard-related overrides that trigger when a bootstrap-table search  %>
@@ -57,8 +64,12 @@
          $('.search-input').each(function (index) {
            $(this).attr('aria-label', 'Search');
          });
+
+         $(".select-institution").select2({
+           language: "<%= params[:locale] %>",
+           width: "100%"
+          });
        });
      </script>
-
   </body>
 </html>
--- a/app/presenters/ht_contact_presenter.rb
+++ b/app/presenters/ht_contact_presenter.rb
@@ -22,7 +22,8 @@ class HTContactPresenter < ApplicationPresenter
   end
 
   def edit_inst_id(form:)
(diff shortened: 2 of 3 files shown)


respinos:
Technically HTDL uses `selectwoo`, because at one time that was the more accessible version.

moseshll:
Hard to say for sure whether `select2` or `selectwoo` looks more in danger of becoming abandonware.

app/views/layouts/application.html.erb
```diff
@@ -9,6 +9,13 @@
     <%= stylesheet_link_tag    'application', media: 'all', 'data-turbolinks-track': 'reload' %>
     <%= javascript_include_tag 'application', 'data-turbolinks-track': 'reload' %>
 
+    <% # Select2 (select2.org) main distro, locale files, CSS %>
+    <%= javascript_include_tag "https://cdn.jsdelivr.net/npm/select2@4.1.0-rc.0/dist/js/select2.min.js" %>
```
mwarin, line 13: Are we OK with using remote-hosted scripts? 
Would cdn.jsdelivr.net take OTIS down if it crashed (and user did not have it cached) ?

respinos:
> 

Indeed. I was hoping the `datalist` would be the modern equivalent, but it only deals with text, not value/label pairs.

app/views/layouts/application.html.erb
respinos: LIT is using `unpkg.com` as a CDN, although we're exploring ways to cascade from `unpkg.com` failures (which did happen recently).

app/views/layouts/application.html.erb
```diff
@@ -9,6 +9,13 @@
...
+    <% I18n.available_locales.each do |locale| %>
+      <%= javascript_include_tag "https://cdn.jsdelivr.net/npm/select2@4.1.0-rc.0/dist/js/i18n/#{locale.to_s}.js" %>
```
mwarin, line 15: I am still seeing "no results found" in en despite locale=jp
image

app/views/layouts/application.html.erb
moseshll: We depend on remote Bootstrap and Bootstrap-table, so this one going poof is no worse than those two. App would totally look like crap but it would still function. Ckeditor (the inline e-mail editor) is also remote, and that one would actually cripple OTIS if it disappeared. I fear this is just the Achilles heel of the state of the art in web apps.

app/views/layouts/application.html.erb
moseshll: Do you see the locale files (en.js, ja.js) coming down the pipe when you Inspect via the network tab? (I'm using Firefox dev tools) here?

moseshll: Oh. Duh. You're not localized at all.

moseshll: `locale=ja`

mwarin (approved):
The basic functionality of searching for institutions works as advertised. 
I was a little confused by the localization stuff and tried to express that confustion in comments, but I see no reason not to approve.

app/views/layouts/application.html.erb
moseshll: I always though `ja` should be Javanese and `jp` should be 日本語

mwarin: OK, yeah that sure made a difference. Comment withdrawn.

moseshll: My bad

moseshll:
@respinos which selectWoo repo are we using in the WAYF? jadu/selectWoo or woocommerce/selectWoo (or something else)? May be another opportunity for a shared repo depending on how and where the files are being stored.

respinos:
We've had https://github.com/woocommerce/selectWoo 5.0.0. vendor-ed in `common` ¯\_(ツ)_/¯  You probably should stick with the CDN; as you noted we really shouldn't be building hard dependencies on this widget.

PR #141, opened 2022-03-08
HT-3362 Registration e-mail subject line should be editable

moseshll:
- Approval request UI updated for feature parity.
- ckeditor UI language synchronized with OTIS locale.
- Borked-up 'de' locale setting for ckeditor (left over from big localization experiment) fixed.
- Use `I18n.locale` instead of `params[:locale]`.

### DIFF
--- a/app/controllers/ht_approval_requests_controller.rb
+++ b/app/controllers/ht_approval_requests_controller.rb
@@ -39,7 +39,8 @@ class HTApprovalRequestsController < ApplicationController
 
   def update
     begin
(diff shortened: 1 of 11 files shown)


aelkiss (approved):
Looks straightforward enough. We'll want to make sure that we have a real default subject in place later -- perhaps we should use the current default subject from `ApprovalRequestMailer` and `RegistrationMailer` and then remove those?

moseshll:
@aelkiss As far as I can tell the defaults in the UI are coming from the mailer class methods which should be the SPOT for those. Unless I missed something, or misunderstand. (And I'm happy to leave those alone I18n-wise because we're not doing localization for the recipient.)

aelkiss:
You're right -- I was confused about the label vs. the value.

PR #143, opened 2022-03-10
HT-1936 registration identity details

moseshll:
- Use MaxMind Ruby interface to version 2 city lite database.
- For development and test, use fake DB from MaxMind GitHub repo.
- *I probably have a bogus path for the GeoIP DB for production -- it is a placeholder* (see `config/initializers/services.rb`)


app/presenters/ht_registration_presenter.rb
```diff
@@ -52,13 +63,33 @@ def show_dsp_name
...
   end
 
-  def show_env
-    return unless self[:env].present?
+  def show_detail_display_name
+    self[:env]["HTTP_X_SHIB_DISPLAYNAME"]
+  end
+
+  def show_detail_edu_person_principal_name
+    self[:env]["HTTP_X_SHIB_EDUPERSONPRINCIPALNAME"]\
```
aelkiss, line 71: Trailing '\'?

aelkiss:
This looks good from what I see, but enough is mocked or doesn't work in dev that I really want to try it in testing.

This might be a good opportunity to:

1) merge #144 
2) look at updated github actions that allow us to manually deploy any branch to testing or staging -- see e.g. https://github.com/mlibrary/patron_account/blob/main/.github/workflows/manual-deploy-testing.yaml and compare to our deploy-testing workflow

app/presenters/ht_registration_presenter.rb
moseshll: Will be fixing that with a force-push momentarily.

config/initializers/services.rb
```diff
@@ -36,3 +37,11 @@ def assign_db(lhs, rhs)
 end
 
 Services.register(:whois) { Whois::Client.new }
+
+geoip_path = if Rails.env.production?
+  "/htprep/geoip/GeoLite2-City.mmdb"
```
aelkiss, line 42: We already have the per-environment configuration facility, and the ability to override it however we like, so we don't need to worry about this here:

```ruby
Services.register(:geoip)  { MaxMind::GeoIP2::Reader.new(Otis.config.geoip_path) }
```

and in `config/settings.yml` add:

```ruby
geoip_path: <%= Rails.root.join("vendor/geoip/GeoIP2-City-Test.mmdb" %>
```

moseshll: After some further investigation, looks like Ettin does not like ERB interpolations that reference `Rails`. I can pull that `geoip_path` logic out into `application.rb` which is more config-centric. Maybe stick the production `htprep` path into `settings.yml` as a default (or do we want to get that from `ENV` only?). But alas I can propose nothing of real elegance.

aelkiss: I see -- probably an issue with load/initialization order.

I would say that since in dev we're running in a container, we can just default to the full path:

```yaml
geoip_path: /usr/src/app/vendor/geoip/GeoIP2-City-Test.mmdb
```

moseshll: After a randomly failing test (weird) looks like that works.

app/presenters/ht_registration_presenter.rb
```diff
@@ -52,13 +63,33 @@ def show_dsp_name
...
-    fields = JSON.parse self[:env]
-    return unless fields.any?
+  def show_detail_geoip
+    geoip = Services.geoip.city ip_address
+    [geoip.country.name, geoip.most_specific_subdivision.name, geoip.city.name].join(" — ")
+  rescue => _e
+    ""
+  end
+
+  def show_detail_organizations
```
aelkiss, line 81: I don't think the whois parser is going to work for reverse lookups, sadly.

I think we do definitely want to be showing the value for HTTP_X_SHIB_IDENTITY_PROVIDER, mapped to an institution name through the institutions table.

app/presenters/ht_registration_presenter.rb
```diff
@@ -4,7 +4,12 @@ class HTRegistrationPresenter < ApplicationPresenter
   ALL_FIELDS = %i[
     dsp_name dsp_email dsp_date inst_id jira_ticket
     auth_rep_name auth_rep_email auth_rep_date
-    contact_info mfa_addendum sent received finished ip_address env
+    contact_info mfa_addendum sent received finished ip_address
+  ].freeze
+
+  DETAIL_FIELDS = %i[
+    detail_edu_person_principal_name detail_display_name detail_geoip
```
aelkiss, line 11: I would like to also see the mail and eduPersonScopedAffiliation attributes mapped here.

app/presenters/ht_registration_presenter.rb
```diff
@@ -52,13 +63,33 @@ def show_dsp_name
...
+    Services.whois.lookup(ip_address).parser.contacts
+      .map { |c| (c.name || c.organization || "").strip.gsub("\n", " — ") }
+      .join("<br/>")
+  rescue => _e
+    ""
+  end
 
-    ["<pre>", fields.map { |k, v| "<strong>#{k}</strong> #{v}" }, "</pre>"].join "\n"
+  def show_detail_reverse_lookup
+    Resolv.getname ip_address
```
aelkiss, line 90: Doesn't seem to work in test, and we're swallowing the exception so it's not clear why.

moseshll: Should be addressed in the next commit with safety `require "resolv"` and some error handling.

app/presenters/ht_registration_presenter.rb
moseshll: Addressed in next commit

aelkiss:
I was able to do the complete end-to-end process via otis-testing, including:
- creating the registration & getting redirected to the preview
- sending the email
- receiving the email
- using the link & getting confirmation
- reviewing the details
- confirming the user

I think we can go ahead and merge this, but let's wait on deployment until the other pending PRs are merged as well.

PR #144, opened 2022-03-11
HT-3395 make sure all ht containers are multiarch

moseshll:
- ARM/AMD architectures, same as used for `db-image`.
- Slightly jumping the gun on HT-3335, move from DockerHub to GHCR as was done with `db-image`.
- Add `org.opencontainers.image.source` metadata for GHCR.


moseshll:
Force-pushed some file edits + renames in preparation for main -> master

moseshll:
This is not all of HT-3334, I suspect the remainder will spill over into a new branch.

aelkiss (approved):

PR #145, opened 2022-03-16
HT-3335 update otis github actions

moseshll:
- Using mainly `mlibrary/patron_account` as a model for the k8s workflow examples.
- See comment questioning need for multiarch support in `manual-deploy-*`.
- Ignoring for now the massive duplication between `manual-deploy-*` workflows.


.github/workflows/manual-deploy-staging.yaml
```diff
@@ -0,0 +1,80 @@
...
+    - name: Set up Docker Buildx
+      if: ${{ steps.image_check.outputs.image_exists != 'true' }}
+      uses: docker/setup-buildx-action@v1
+    - name: Build and Push
+      if: ${{ steps.image_check.outputs.image_exists != 'true' }}
+      uses: docker/build-push-action@v2
+      with:
+        context: .
+        file: Dockerfile.prod
+        # This (and the QEMU/buildx setup) may not be needed if we only want to get
```
aelkiss, line 66: Right, it's not an issue in k8s; it only comes up when we want to reuse these images for development on the desktop. I think my inclination is just to include multiarch by default to avoid the potential for confusion down the line. If eventually we can avoid duplication across actions/workflows the only reason to avoid it is if it's particularly time-consuming to do so & we're trying to cut down on our github actions CPU usage. I don't think we're particularly close to hitting that limit any time soon.

moseshll: Agreed. That said, be aware multiarch builds take significantly longer than the single-arch "native" builds we've been doing. 18-22 minutes vs the previous 3-ish.

aelkiss: Hmm, OK. I was hoping for just double instead of 7x, but I guess that makes sense if it's via emulation for arm. We can give it a try and see what we think, I guess.

moseshll: C compiler or compiler-adjacent (like `ldconfig`)? I dunno. But it sure smells like QEMU is the time hog. With our `n=2` samples.

aelkiss (commented):
There are a couple places where the image name still needs to be changed to ghcr.io. I'm anxious to try this out for deploy to testing, so I'll just push an update.

PR #148, opened 2022-03-23
DEV-49 allow deletion of sent registration

moseshll:
- Sent registrations can be deleted just like unsent ones.
- Once received or finished the option to delete disappears (another strategy must be used for auto-hiding or removing old data).
- Use `assert_select` in tests as it is more reliable than string matching with path helpers.


aelkiss:
I will give this a try in development.

aelkiss:
Looks good. Verified I could delete a sent registration; verified that delete button no longer appears after registration is received.

aelkiss (approved):

PR #149, opened 2022-03-30
DEV-216 Remove 'Logged In As' from header

moseshll:
- Replace with HathiTrust logo and indication of login status without displaying potentially gibberish user id.


moseshll:
> Looks good to me. Has Jessica reviewed/provided/approved the branding?

No. I am using the same official image (from the Drive folder) that I swapped in for e-mail branding.

aelkiss:
OK -- probably worth a quick check with Jessica to make sure it's the current/correct one then?

PR #152, opened 2022-04-01
DEV-61 E-mail Branding

moseshll:
- Add official HathiTrust horizontal logo to registration and approval request e-mails.


aelkiss (approved):
Looks fine to me; I did not test directly. I would be fine merging and then checking via staging that it works as expected before we release & deploy.

PR #153, opened 2022-04-04
DEV-234: Match behavior from mdp-lib for user id

aelkiss:
This is 100% technical debt; it is worth a discussion to see what we
would need to do to eliminate the special cases and transformation. We
may be farther along than I had thought (see discussion on DEV-234).


test/controllers/ht_registrations_controller_test.rb
```diff
@@ -414,6 +414,40 @@ def setup
     assert_equal "edit", @controller.action_name
   end
 
+  test "finalized user uses downcased shibboleth id" do
```
aelkiss, line 417: @moseshll I added tests here since there weren't any existing tests for `RegistrationMover`, but I could certainly do that if you prefer.

moseshll: Yeah, 100% coverage in `RegistrationMover` in Coveralls but all from controller testing. I'd say create a new `RegistrationMover` test file -- we're sure to need it.

moseshll (changes requested):
As mentioned in the convo, since we expect this to change into a saner logic at some point, it probably makes sense to have the added tests explicitly for RegistrationMover it their own suite. Unless, of course, there turns out to be a dependency on having a controller, or some other kind of Rails weirdness.

moseshll (approved):

aelkiss:
rebased, squashed, force-pushed; will merge once tests pass again.

PR #154, opened 2022-04-06
DEV-42 Jira automation for registration events.

moseshll:
- Add `Otis::JiraClient` class.
- Update credentials to talk to Jira as `lit-cs-ingest`.
- Keep development and test environments "readonly" for the purpose of submitting issue comments.
- Test coverage suffers a bit because no mocks are being used.
- Jira comments are terse in comparison to those from the legacy systems since we're keeping the details accessible within OTIS.


moseshll:
@aelkiss I gave WebMock a shot but it looks like updates have lapsed and it doesn't advertise Ruby 3 compatibility. I couldn't get it to work for URI regexes so tried this `NullJiraObj` approach instead.

app/lib/otis/jira_client.rb
```diff
@@ -0,0 +1,74 @@
+# frozen_string_literal: true
+
+require "jira-ruby"
+
+module Otis
+  class JiraClient
+    # Fake JIRA::Client used when Otis.config.jira.readonly is set.
+    # It just responds to most methods by returning itself.
+    class NullJiraObj
```
aelkiss, line 9: I'd be inclined to move this into the tests and allow it to be injected into `Otis::JiraClient` at constructor time:

```ruby
def initialize(client = connect)
  @client = client
end
```

and then in tests

```ruby
OtisJiraClient.new(client = NullJiraObj.new)
```

test/jira_client_test.rb
```diff
@@ -0,0 +1,21 @@
...
+require "otis/jira_client"
+
+module Otis
+  class JiraClientTest < ActiveSupport::TestCase
+    test "make sure we're read-only" do
+      assert Otis.config.jira.readonly
+    end
+
+    test "comment template test" do
+      comment = Otis::JiraClient.comment template: :registration_sent, user: "nobody@default.invalid"
```
aelkiss, line 13: It doesn't look like this actually connects to anything, so do we even need the null jira object?

moseshll: I've placed the comment templates in the so-called Jira client so they're accessible to both the registration controller and the finalize controller, and semantically they're "stuff we send to Jira". There's no supposition of any sort of HTTP client, it's more of a utility method.

app/lib/otis/jira_client.rb
moseshll: Yes, I like that approach better.

moseshll: Except we also want a null object in development, in case we accidentally use a real Jira ticket number when running through registration scenarios. What if we put an explicit factory in Otis::JiraClient that explicitly uses Rails.env.production? to determine what to return? And dispense with the "readonly" config?

test/jira_client_test.rb
aelkiss: :+1:

app/lib/otis/jira_client.rb
aelkiss: Yes, I think that approach makes sense.

moseshll:
@aelkiss Kinda came full circle on the way I'm doing this, but here we are. I think the important thing is, we're not hitting Jira at all in dev/testing, but we sorta have some test coverage.

aelkiss (approved):
Looks good for now; if/when we move away from Jira we can revisit.

PR #155, opened 2022-04-11
DEV-212 Make registration more like unsubscribing from mailing list

moseshll:
- Still using the "new" controller action, but rendering either "edit" with a form, or "show" with a success message.
- No longer scold the user for following the link twice or reloading the page.
- Awaiting feedback on the content of the three success messages, so leaving this in draft in hope of getting the nontechnical details squared
away on this ticket/PR.
- db/seeds.rb removes old `ht_web.otis_registrations` entries that can cause referential integrity issues with deleted `ht_institutions`.

### DIFF
--- a/app/assets/stylesheets/otis.scss
+++ b/app/assets/stylesheets/otis.scss
@@ -59,3 +59,10 @@ a:visited {
 label a {
   font-weight:400
 }
+
+.complete-registration-link {
+  font-weight: bold;
+  font-family: courier;
+  font-size: 120%;
+  color: #d65d22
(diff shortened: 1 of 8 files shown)


moseshll:
@mwarin I had to force push because of the rebase but the typo fix should be integrated now. I still need to do one more change to fix the styling on the e-mail finalize link we looked at -- I'll do it on this branch, but later in the day.

moseshll:
E-mail styling fix added & rebased. This should cover the issues we encountered.

PR #157, opened 2022-04-14
DEV-50 Update registration form fields to match new registration form.

moseshll:
- Rename `otis_registrations` fields starting with `dsp_` to `applicant_`.
- Add 30- and 180-day expiration periods introduced in CAA registration form.
- Add `hathitrust_authorizer` field for CAA users.
- Add `role` field with limited set of ATRS + CAA roles (and note that this smaller set will need to be reconciled with what's in `ht_users`).

### DIFF
--- a/app/controllers/finalize_controller.rb
+++ b/app/controllers/finalize_controller.rb
@@ -7,7 +7,7 @@
 class FinalizeController < ApplicationController
   def new
     @token = params[:token]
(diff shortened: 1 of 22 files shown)


moseshll:
Still need to address `atrs` vs `ssdproxy` role on this branch so there will be one more force push.

mwarin (approved):
Looks OK to me.

PR #158, opened 2022-04-14
DEV-46 Review registration and renewal emails to ensure alignment wit…

moseshll:
…h service description

- Update registration e-mail based on input from HathiTrust staff.

### DIFF
--- a/app/views/layouts/mailer.html.erb
+++ b/app/views/layouts/mailer.html.erb
@@ -1,12 +1,17 @@
 <!DOCTYPE html>
-<html>
+<html lang="en">
   <head>
+    <title>HathiTrust Elevated Access</title>
     <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
     <style>
-      /* Email styles need to be inline */
+      .complete-registration-link {
+        font-weight: bold;
+        font-family: courier;
+        font-size: 120%;
+        color: #d65d22
+      }
     </style>
   </head>
-
   <body>
     <%= yield %>
   </body>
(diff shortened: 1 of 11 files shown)


app/views/shared/_registration_link.html.erb
```diff
@@ -1,4 +1,4 @@
-<a href="<%= finalize_url(@registration.token, host: @base_url) %>"
+<a href="<%= finalize_url(@registration.token, host: @base_url, locale: nil) %>"
    style="font-weight:bold; font-family:courier; font-size:120%; color:#d65d22">
```
mwarin, line 2: I'd pick `class="..."`  over `style="..."` .

mwarin (approved):
Looks OK.

app/views/layouts/mailer.html.erb
```diff
@@ -1,12 +1,17 @@
 <!DOCTYPE html>
-<html>
+<html lang="en">
```
mwarin, line 2: Should be `<%= I18n.locale %>` maybe?

moseshll: It's a W3C recommendation to have the tag, so it shuts up the pedantic warnings :-)
But actually, no we don't want our emails to (necessarily) be in the same language as the UI language, they're orthogonal. It's a can I/we deliberately kicked down the road when I did the wacko I18n branch. One that will prolly be painful when we revisit it.

moseshll: Actually it won't be that painful. When we have multiple languages for our outgoing e-mails we will have a "flag" menu in the preview page. How we're to know which language to send is anyone's guess, however.

mwarin: `email_lang = I18n.locales.sample`

PR #160, opened 2022-04-18
DEV-243 OTIS tests sporadically fail

moseshll:
- Add `require "ip_restriction"` to test file to prevent occasional uninitialized constant errors.


mwarin (approved):
if only all code reviews were this simple

PR #161, opened 2022-05-02
Update CVE-2022-27777 and CVE-2022-22577

moseshll:
- Dependabot fails to automatically create a patch for our particular config.
- Basically just update everything Rails-related from 6.1.4.7 to 6.1.5.1.


mwarin (approved):
Looks fine and the tests are green, so I got nothing more to say.

PR #163, opened 2022-05-05
DEV-261 Include ATRS instruction link in OTIS registration e-mail

moseshll:
- New language for registration e-mail vetted by Melissa S.


mwarin (approved):
I trust that this shows up when it is supposed to, being a small text change to something that has already been reviewed.

PR #165, opened 2022-05-06
DEV-265 OTIS vs WHOIS

moseshll:
- Move WHOIS query from registration show page view to presenter.
- Add WHOIS exception handling and localized error message.
- Add ability to force WHOIS testing stub to throw exception.
- Sneak current locale into `<html>` tag in app layout (to make HTML validators happy).

### DIFF
--- a/app/presenters/ht_registration_presenter.rb
+++ b/app/presenters/ht_registration_presenter.rb
@@ -32,6 +32,14 @@ class HTRegistrationPresenter < ApplicationPresenter
     self.class::DETAIL_FIELDS
   end
 
+  def whois_block
+    begin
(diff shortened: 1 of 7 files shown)


moseshll:
@mwarin if you decide you want to run this one to try it out, I found the easiest place to cause WHOIS to blow up is to swap something bogus for the call to the WHOIS service, like the zero division in

```
def whois_block
    begin
      1/0
    rescue => e
      I18n.t("errors.whois", err: e.to_s)
    end.yield_self { |s| "<pre>#{s}</pre>" }.html_safe
  end
```

moseshll:
Please oh please let this be the last registration hotfix.

mwarin (approved):

PR #167, opened 2022-05-24
DEV-54 Daily Digest E-mails

moseshll:
- Add DailyDigestMailer mailer class with Otis::DailyDigest "helper" class.
- Add minimalist Rake task to be configured via ht_tanka.
- Add some utility facets to HTApprovalRequest and HTRegistration to make queries simpler.
- Mailer preview in test/ directory.
- Add some CSS table classes in the default mailer layout. Keeping them in assets would likely require another gem.
- Prevent ht_user primary key collision in factories.rb for testing.
- Seed database with registrations in different stages of completion.
- Quick fix for registration landing page typo identified in HTS-151344.


app/lib/otis/daily_digest.rb
```diff
@@ -0,0 +1,25 @@
...
+      digest = DailyDigest.new
+      DailyDigestMailer.with(digest: digest).daily_digest_email.deliver_now
+    end
+
+    def initialize
+      @ready_registrations = HTRegistration.ready
+      @expired_registrations = HTRegistration.expired
+      @ready_approval_requests = HTApprovalRequest.approved.not_renewed
+      @expired_approval_requests = HTApprovalRequest.expired
+      @expiring_users = HTUser.active.select do |user|
```
aelkiss, line 20: Probably fine given the relatively small number of users and that this isn't being run interactively, but I do wonder if it would be worth thinking about how to encode this as a SQL query so we don't have to load as many users.

db/seeds.rb
```diff
@@ -134,6 +133,16 @@ def create_ht_registration(inst_id)
     mfa_addendum: [true, false].sample,
     contact_info: Faker::Internet.email
   )
+  if rand < 0.5
```
aelkiss, line 136: I'd consider making these "sent" and "received" traits on the `ht_registration` factory and then doing something like

```ruby
[build(:ht_registration), build(:ht_registration,:sent), build(:ht_registration,:received)].sample.save!
```

aelkiss (approved):
I didn't try it out, but the tests and logic look good to me. I did add a couple comments with areas to consider changing, but I don't think it needs to block merging.

db/seeds.rb
moseshll: I think I'm going to disagree on this one based on what I understand about FactoryBot: we are probably creating factory objects that are too completely specified, and tests should manipulate factory objects that have only the fields that are `NOT NULL` or are germane to the test at hand. Currently our seeding/factorying practices are eerily similar but I think the push should not be to merge them.

aelkiss: Fine by me.

moseshll:
I'll force-push the commit with the added expiring users scope and merge.

PR #169, opened 2022-06-02
DEV-286 OTIS should not allow new registrations for existing users

moseshll:
- Add `RegistrationMover.user_exists?` utility called by registrations controller.
- This seemed like the logical place to put it since that class concerns itself with both registrations and users.

### DIFF
--- a/test/controllers/ht_registrations_controller_test.rb
+++ b/test/controllers/ht_registrations_controller_test.rb
@@ -230,6 +230,15 @@ class HTRegistrationsControllerCreateTest < ActionDispatch::IntegrationTest
     assert_equal 0, HTRegistration.count
     assert_not_empty flash[:alert]
   end
+
+  test "refuses to create registration for someone already in ht_users" do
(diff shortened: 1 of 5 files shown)


aelkiss (approved):
Looks good -- concise, and the test clearly covers the situation we care about.

test/controllers/ht_registrations_controller_test.rb
```diff
@@ -230,6 +230,16 @@ def setup
     assert_equal 0, HTRegistration.count
     assert_not_empty flash[:alert]
   end
+
+  test "refuses to create registration for someone already in ht_users" do
+    user = FactoryBot.create(:ht_user)
+    params = FactoryBot.build(:ht_registration, applicant_email: user.email)
```
aelkiss, line 236: If memory serves I think you can use `FactoryBot.attributes_for` for this situation

moseshll: That is correct. And more concise, as it turns out. Change made.

PR #170, opened 2022-06-16
DEV-54 fix for OTIS daily digest

moseshll:
- Rake task uses ENV['RAKE_DEFAULT_URL_HOST'] for default URL host in ActionMailer config.
- Some peculiar shenanigans required for testing.
- No changes to config for OTIS as a whole.


aelkiss (approved):
Agreed kind of ugly for the test, but makes sense.

PR #173, opened 2022-06-21
Use correct mlibrary/deploy-to-kubernetes@v2 'registry' instead of 'g…

moseshll:
…ithub' keys.

### DIFF
--- a/.github/workflows/deploy-production.yaml
+++ b/.github/workflows/deploy-production.yaml
@@ -15,8 +15,8 @@ jobs:
     - name: Deploy to Production
       uses: mlibrary/deploy-to-kubernetes@v2
       with:
-        github_username: ${{ github.actor }}
-        github_token: ${{ secrets.GITHUB_TOKEN }}
+        registry_username: ${{ github.actor }}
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-production.yaml
```diff
@@ -15,8 +15,8 @@ jobs:
     - name: Deploy to Production
       uses: mlibrary/deploy-to-kubernetes@v2
       with:
-        github_username: ${{ github.actor }}
-        github_token: ${{ secrets.GITHUB_TOKEN }}
+        registry_username: ${{ github.actor }}
+        registry_token: ${{ secrets.GITHUB_TOKEN }}
         image: hathitrust/otis:${{ github.event.inputs.tag }}
```
aelkiss, line 20: Also need to update this to `ghcr.io/hathitrust/otis`...

aelkiss:
Needs the image updated. You can try this by using the action and specifying "use workflow from branch" `deploy_production_cronjob_part_deux` with the same release

aelkiss (approved):

PR #175, opened 2022-07-01
DEV-314 OTIS does not do anything with MFA Addendum

moseshll:
- Create new user with MFA enabled if registration MFA Addendum is set.
- Check for existing registration with addendum when validating `ht_user.mfa`.
- Note that the latter depends on old registrations hanging around in order to license `ht_user.mfa` when it conflicts with our institution record.

### DIFF
--- a/app/lib/otis/registration_mover.rb
+++ b/app/lib/otis/registration_mover.rb
@@ -15,13 +15,14 @@ module Otis
     def ht_user
       return @ht_user unless @ht_user.nil?
 
+      institution = HTInstitution.find(@registration.inst_id)
       @ht_user = HTUser.new(userid: userid,
         email: @registration.applicant_email, displayname: @registration.applicant_name,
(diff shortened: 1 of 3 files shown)


moseshll:
That's weird. I can't reproduce that issue here. I was using one of the seeded addendum=1, mfa=0 users (also tried it with addendum=1, mfa=1). Had to splice in a fake user name in RegistrationMover since when using the dev-only URL in the flash message no ENV gets populated (could fake up that data but it's more dev-only stuff woven into the production code). Can you send/attach the Rails console params for the ht_user update, or the HTLogs entry?

aelkiss:
I might have something in a weird state -- I'll try cleaning everything out and trying again.

aelkiss (approved):
Tested locally, works as expected:

- verified MFA addendum user has ip restriction set to any
- verified entity ID is populated

PR #179, opened 2022-07-25
DEV-351 OTIS ht_institutions.us display broken

moseshll:
- Add reference to flag-icons-rails in app SCSS to restore US flag to ht_institutions.us.
- Replace deprecated config.web_console.whitelisted_ips (unrelated but necessary to do eventually).


moseshll:
Checks out in testing environment, going ahead with merge.

PR #180, opened 2022-08-04
DEV-319 Register IDP Change

moseshll:
*As a CAA user, I want to re-gain access when my identity as provided by my IdP has changed so that I can continue to use the service.*
- Allow HTRegistration to be created even if it shares an e-mail with an existing user.
  - Add localized "Existing User" badge to registration email field (on show page) as a warning that this is a re-registration.
- `RegistrationMover` updates existing users where appropriate instead of always creating anew.
- Condition availability of "Create User" (and new alternate "Update User") button on `finished` status of registration.
- Add a warning + no-op if the operator somehow manages to submit "Create/Update User" twice.
- QoL improvements for the devs
  - Remove that *gottverdammt* pid file when running `web` service.
  - Create some fake registrations with `env` values when seeding DB.
  - Update Faker gem for `Faker::Internet.base64` to produce high-quality Shib gobbledygook.
- HTInstitution factory seems to be able to produce `inst_id` collisions? Try to mitigate with `Faker...unique`.

### DIFF
--- a/app/lib/otis/registration_mover.rb
+++ b/app/lib/otis/registration_mover.rb
@@ -10,9 +10,8 @@ module Otis
       return @ht_user unless @ht_user.nil?
 
       institution = HTInstitution.find(@registration.inst_id)
-      @ht_user = @registration.existing_user || HTUser.new
-      @ht_user = HTUser.new(userid: userid,
-        email: @registration.applicant_email, displayname: @registration.applicant_name,
+      @ht_user = @registration.existing_user || HTUser.new(email: @registration.applicant_email)
+      @ht_user.update(userid: userid, displayname: @registration.applicant_name,
         inst_id: @registration.inst_id, identity_provider: institution.entityID,
         approver: @registration.auth_rep_email, authorizer: authorizer,
         expire_type: @registration.expire_type,
--- a/app/lib/otis/registration_mover.rb
+++ b/app/lib/otis/registration_mover.rb
@@ -2,12 +2,6 @@ module Otis
   # Responsible for creating a new HTUser, populating it with the data in
   # an HTRegistration, and saving.
   class RegistrationMover
-    UMICH_IDP = "https://shibboleth.umich.edu/idp/shibboleth"
-
-    def self.user_exists?(registration)
-      HTUser.where(email: registration.applicant_email).exists?
-    end
-
     def initialize(registration)
       @registration = registration
     end
@@ -16,6 +10,7 @@ module Otis
       return @ht_user unless @ht_user.nil?
 
       institution = HTInstitution.find(@registration.inst_id)
+      @ht_user = @registration.existing_user || HTUser.new
       @ht_user = HTUser.new(userid: userid,
         email: @registration.applicant_email, displayname: @registration.applicant_name,
         inst_id: @registration.inst_id, identity_provider: institution.entityID,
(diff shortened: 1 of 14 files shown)


app/lib/otis/registration_mover.rb
```diff
@@ -4,10 +4,6 @@ module Otis
   class RegistrationMover
     UMICH_IDP = "https://shibboleth.umich.edu/idp/shibboleth"
```
mwarin, line 5: UMICH_IDP could be moved out to some conf file, no?

aelkiss (approved):
Looks great at first glance. I will give it a try locally and make sure I can follow through the workflow. We may be getting to the point where it would be helpful to have some feature tests w/ capybara -- if nothing else, they can be helpful in terms of understanding the expected workflows.

docker-compose.yml
```diff
@@ -12,7 +12,7 @@ services:
     depends_on:
       - mariadb-dev
       - mariadb-test
-    command: bash -c "bin/init_dev_db.sh && bundle exec rails s -b 0.0.0.0"
+    command: bash -c "rm -f /usr/src/app/tmp/pids/server.pid && bin/init_dev_db.sh && bundle exec rails s -b 0.0.0.0"
```
mwarin, line 15: nice

mwarin (approved):
looks good

app/lib/otis/registration_mover.rb
moseshll: Yes indeed. Will force push a patch.

moseshll:
@respinos This branch is deployed to test (https://test.babel.hathitrust.org/otis) if you want to try it out. Alas it is a multistep process so let me know if you want to do so on a call.

app/lib/otis/registration_mover.rb
```diff
@@ -16,6 +10,7 @@ def ht_user
       return @ht_user unless @ht_user.nil?
 
       institution = HTInstitution.find(@registration.inst_id)
+      @ht_user = @registration.existing_user || HTUser.new
```
respinos, line 13: Does this `@ht_user` not immediately get overwritten in the next line?

respinos: @moseshll when I create a new registration with an existing email address, I get the "existing user" badge, but when I try to finish the registration I get a warning about "E-mail has already been taken".

moseshll: That's bonkers. It looks like the thing got munged when I rebased, because there was quite a bit of clowning around that had to be done in the registration mover -- Rails gets irked when you try to change a primary key and just discards the new value.

Clearly the testing is faulty as well, because this never should have gotten by.

moseshll: Fixed logic and added some sanity checks to tests.

moseshll: @respinos Update deployed to testing. Is your testing data in a state to try a resubmit?

respinos: @moseshll Now I get `User ID can't be blank and Multi-Factor must be blank` --- but if it was attaching to the existing `ht_user` then there'd be a username, right?

moseshll: Those are heavily dependent on the Shib `otis_registrations.env` picked up by the "finalize" process. Leaving aside uniqname weirdness, user ID will be from HTTP_X_REMOTE_USER and it would overwrite any existing value for `ht_users.userid`. The multi-factor error is a bit more concerning. I take it you are running this locally? That bit should be taking care of itself so long as the institution the new user is attaching to is well-formed.

PR #181, opened 2022-08-10
DEV-218 Authentication data matching for registration

moseshll:
- Should look at this with MS in a test environment.

### DIFF
--- a/app/presenters/ht_registration_presenter.rb
+++ b/app/presenters/ht_registration_presenter.rb
@@ -19,6 +19,7 @@ class HTRegistrationPresenter < ApplicationPresenter
   READ_ONLY_FIELDS = %i[sent received finished ip_address env].freeze
   JIRA_BASE_URL = URI.join(Otis.config.jira.site, "/jira/", "browse/").to_s.freeze
   FIELD_SIZE = 45
+  VALID_AFFILIATIONS = %r{^(faculty|staff|member|employee)}
 
   BADGES = {
     sent: Otis::Badge.new("activerecord.attributes.ht_registration.sent", "label-info"),
@@ -26,7 +27,10 @@ class HTRegistrationPresenter < ApplicationPresenter
     finished: Otis::Badge.new("activerecord.attributes.ht_registration.finished", "label-success"),
     institution_static_ip: Otis::Badge.new("activerecord.attributes.ht_registration.institution.static_ip", "label-danger"),
     institution_mfa: Otis::Badge.new("activerecord.attributes.ht_registration.institution.mfa", "label-success"),
-    existing_user: Otis::Badge.new("activerecord.attributes.ht_registration.email.existing_user", "label-warning")
+    existing_user: Otis::Badge.new("activerecord.attributes.ht_registration.email.existing_user", "label-warning"),
+    ok: Otis::Badge.new("activerecord.attributes.ht_registration.detail.ok", "label-success"),
+    mismatch: Otis::Badge.new("activerecord.attributes.ht_registration.detail.mismatch", "label-danger"),
+    questionable: Otis::Badge.new("activerecord.attributes.ht_registration.detail.questionable", "label-warning")
   }.freeze
 
   def detail_fields
@@ -79,11 +83,19 @@ class HTRegistrationPresenter < ApplicationPresenter
   end
 
   def show_detail_edu_person_principal_name
-    env["HTTP_X_SHIB_EDUPERSONPRINCIPALNAME"]
+    name = env["HTTP_X_SHIB_EDUPERSONPRINCIPALNAME"] || ""
+    # Check if principal name or email matches the given DSP email.
+    # Not all IDPs send this info, so no full-blown warning.
+    badge = name == applicant_email ? :ok : :questionable
+    name + " " + BADGES[badge].label_span
   end
 
   def show_detail_email
-    env["HTTP_X_SHIB_MAIL"]
+    email = env["HTTP_X_SHIB_MAIL"] || ""
+    # Check if principal name or email matches the given DSP email.
+    # Not all IDPs send this info, so as with eppn no need for full-blown warning.
+    badge = email == applicant_email ? :ok : :questionable
+    email + " " + BADGES[badge].label_span
   end
 
   def show_detail_geoip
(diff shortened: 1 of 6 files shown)


app/presenters/ht_registration_presenter.rb
```diff
@@ -19,14 +19,18 @@ class HTRegistrationPresenter < ApplicationPresenter
   READ_ONLY_FIELDS = %i[sent received finished ip_address env].freeze
   JIRA_BASE_URL = URI.join(Otis.config.jira.site, "/jira/", "browse/").to_s.freeze
   FIELD_SIZE = 45
+  VALID_AFFILIATIONS = %r{^(faculty|staff|member|employee)}
+  AUTH_EXCEPTIONS = ["nypl"]
```
aelkiss, line 23: Consider moving these to configuration

app/presenters/ht_registration_presenter.rb
```diff
@@ -79,11 +83,17 @@ def show_detail_display_name
   end
 
   def show_detail_edu_person_principal_name
-    env["HTTP_X_SHIB_EDUPERSONPRINCIPALNAME"]
+    name = env["HTTP_X_SHIB_EDUPERSONPRINCIPALNAME"] || ""
+    # Check if principal name or email matches the given DSP email
+    badge = name == applicant_email ? :ok : :mismatch
```
aelkiss, line 88: I don't know if we necessarily want to convey "danger" here -- they aren't necessarily expected to match, but it's worth checking that it at least makes sense. It's also not unexpected for this header to be missing (not all identity providers provide it)

app/presenters/ht_registration_presenter.rb
```diff
@@ -79,11 +83,17 @@ def show_detail_display_name
...
+    name + " " + BADGES[badge].label_span
   end
 
   def show_detail_email
-    env["HTTP_X_SHIB_MAIL"]
+    email = env["HTTP_X_SHIB_MAIL"] || ""
+    # Check if principal name or email matches the given DSP email
+    badge = email == applicant_email ? :ok : :mismatch
```
aelkiss, line 95: As with eppn - it's not "danger do not approve" if they don't match, but should invite extra scrutiny; likewise not all IdPs will send it, so not really a problem if it's missing entirely

app/presenters/ht_registration_presenter.rb
```diff
@@ -94,8 +104,15 @@ def show_detail_geoip
...
 
   def show_detail_identity_provider
-    entity = env["HTTP_X_SHIB_IDENTITY_PROVIDER"]
-    ERB::Util.html_escape(HTInstitution.where(entityID: entity).first&.name || entity)
+    idp = env["HTTP_X_SHIB_IDENTITY_PROVIDER"] || ""
+    # Check if identity provider matches the institution picked on the signup form
+    inst = HTInstitution.where(entityID: idp).first
+    badge = ""
+    unless self.class::AUTH_EXCEPTIONS.member?(inst_id)
+      badge_id = idp == ht_institution.entityID ? :ok : :mismatch
```
aelkiss, line 112: This one is definitely worth the "danger do not approve" label if they don't match

app/presenters/ht_registration_presenter.rb
```diff
@@ -105,7 +122,29 @@ def show_detail_reverse_lookup
   end
 
   def show_detail_scoped_affiliation
-    env["HTTP_X_SHIB_EDUPERSONSCOPEDAFFILIATION"]
+    affiliation = env["HTTP_X_SHIB_EDUPERSONSCOPEDAFFILIATION"] || ""
+    # Check if scoped affiliation is faculty@, staff@, member@, or employee@
+    # AND matches the allowed affiliations for the IdP the user logged in with
+    badge = ""
+    unless self.class::AUTH_EXCEPTIONS.member? inst_id
+      badge_id = scoped_affiliation_valid? && scoped_affiliation_match? ? :ok : :mismatch
```
aelkiss, line 130: Likewise I think the danger label is appropriate here

aelkiss (commented):
I'll give it a try in dev. Tests look good; comments are primarily about how stringent a warning to display for certain kinds of mismatches.

app/presenters/ht_registration_presenter.rb
moseshll: Getting the affiliations regex into config requires YAML magic or `eval` and I'm not fired up about either of these. But the exceptions list seems to fit nicely.

aelkiss: :+1: The valid affiliations seems less likely to change; if we find that we need to update them with any frequency we can see what we can do.

moseshll:
Downgraded eppn and mail warnings to "questionable" label and non-red-alert tint.

app/presenters/ht_registration_presenter.rb
```diff
@@ -105,7 +124,29 @@ def show_detail_reverse_lookup
...
+
+    affiliations = env["HTTP_X_SHIB_EDUPERSONSCOPEDAFFILIATION"].downcase.split(";")
+    affiliations.any? { |affiliation| self.class::VALID_AFFILIATIONS.match? affiliation }
+  end
+
+  def scoped_affiliation_match?
+    return false unless env["HTTP_X_SHIB_EDUPERSONSCOPEDAFFILIATION"].present?
+
+    affiliations = env["HTTP_X_SHIB_EDUPERSONSCOPEDAFFILIATION"].downcase.split(";")
+    affiliations.any? { |affiliation| ht_institution.allowed_affiliations.downcase.include? affiliation }
```
aelkiss, line 149: I suspect this is why we're seeing a mismatch in testing -- `ht_institution.allowed_affiliations` is a regex

moseshll: Curses, those pesky DB regexes.

test/presenters/ht_registration_presenter_test.rb
```diff
@@ -89,3 +89,82 @@ def setup
     assert_match "/ht_registrations/#{reg.id}", reg.cancel_path
   end
 end
+
+class HTRegistrationPresenterENVTest < ActiveSupport::TestCase
+  def setup
+    @inst = create(:ht_institution, inst_id: "test",
+      allowed_affiliations: "faculty@default.invalid;staff@default.invalid")
```
aelkiss, line 96: We would normally expect this to be `^(faculty|staff)@default.invalid`

moseshll: Aaron Testuser Elkiss now shows OK badge in testing with latest patch.

aelkiss (approved):
Looks good on testing now. Merging.

PR #183, opened 2022-09-02
Update github actions workflows

aelkiss:
- Use shared actions from github.com/hathitrust/github_actions
- Use coveralls

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,5 @@
+log
+tmp
+db/*.log
+kube
+coverage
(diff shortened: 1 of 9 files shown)


aelkiss:
With the feature specs, we now have two different sets of specs running and both (probably) write coverage information and stomp on each other. We should resolve that so that either we have two different sets of coverage or that one (probably the integration spec) doesn't run coverage. Coveralls probably supports this but the documentation is not great. (Regarding other coverage services, I've looked a bit at codecov and it is generally much better documented and looks nicer, but it doesn't have a currently-functional perl uploader.)

aelkiss:
Yeah, coveralls now sees coverage at 44% (presumably from the system tests) whereas `docker-compose run test` reports 95%

aelkiss:
This looks better on coveralls: https://coveralls.io/builds/52438337

PR #184, opened 2022-10-04
DEV-438 update otis to use support@hathitrust.org

moseshll:
- Move support email to settings and interpolate into templates.
- Fix system test with Capybara ambiguous element error.

### DIFF
--- a/app/views/shared/_approval_request_body.html.erb
+++ b/app/views/shared/_approval_request_body.html.erb
@@ -17,6 +17,6 @@ information.<br/></p>
 
 <p>If you are no longer supervising the staff listed below or if you need to add
 or change staff with elevated access, please email
-<a href="mailto:feedback@issues.hathitrust.org">feedback@issues.hathitrust.org</a>.
+<a href="<%= Otis.config.support_email %>"><%= Otis.config.support_email %></a>.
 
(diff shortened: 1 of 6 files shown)


moseshll:
@mwarin The Capybara error referenced above made me nervous so I'm requesting a sanity check review. I'd kinda like to see this pass the normal and the system tests (`docker-compose run test` and `docker-compose run system-test`) with no weirdness.

mwarin (commented):
Would it be worth sticking the whole 
`<a href="<%= Otis.config.support_email %>"><%= Otis.config.support_email %></a>`
into a var, since you are using that more than a few times?

mwarin:
`docker-compose run system-test` did not give any errors the first time I ran it, but the second time it threw a `ActionController::RoutingError`.

`docker-compose run test` ran fine, but I am seeing more than a few skipped tests (~20). I would recommend seeing if you could unskip or removing (some of) them?

mwarin:
Actually, a test that github ran failed but, coveralls covered the exit code. 
https://github.com/hathitrust/otis/actions/runs/3185370830/jobs/5194870574#step:8:38

mwarin (approved):
I approve these changes, but acknowledge that other issues relating to the capybara tests where uncovered and should be dealt with separately.

PR #186, opened 2022-10-07
DEV-448 OTIS system tests fail silently

moseshll:
- Bail out of `bin/run_system_tests.sh` with nonzero exit status instead of running cleanup code.
    - Fix approval request URL extraction in system tests when there are multiple URLs.

### DIFF
--- a/bin/run_system_tests.sh
+++ b/bin/run_system_tests.sh
@@ -6,4 +6,5 @@ bin/wait-for mariadb-test:3306
 bin/wait-for chrome-server:4444
 bundle exec rake db:seed
 bundle exec rake test:system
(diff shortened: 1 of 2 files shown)


moseshll:
@aelkiss I knocked the coverage pass/fail point down to 50% in the Coveralls settings for this repo since the integration portion was failing unnecessarily with the defaults (yes, we know our integration tests are incomplete, thanks).

mwarin (approved):
Looks good.

aelkiss:
Sounds fine. I think there is a way to get coveralls to consider the coverage for multiple sets of tests, but we can look at that at some future point in time.

PR #190, opened 2022-11-18
DEV-512 HTContact index fails if institution does not exist

moseshll:
- This is a situation encountered in production: contact with inst_id "laurentian" does not correspond to an ht_institution.
- This has been tested in staging as part of another branch (`cp_not_prepared_err`)


aelkiss (approved):
Looks fine. We were using this for ETAS contacts, right? We probably missed deleting the contact when we dropped the institution...

moseshll:
@aelkiss yes, just ETAS contacts.

PR #203, opened 2023-05-08
TTO-66 ssdproxy users should have access=normal

moseshll:
- `RegistrationMover` grants access "normal" to ssdproxy registrations and "total" otherwise.
- Would like confirmation from the reviewer that the business logic is sound.


aelkiss (approved):
Looks correct as I now understand it.

PR #204, opened 2023-08-02
DEV-794 Update logos in mostly-internal applications

moseshll:
- Update logo and email signature
- Move file names and sizes to config/settings.yml.
- Update Dockerfile `netcat` to `netcat-traditional` since the former has disappeared.
- Update `seeds.rb` to purge additional tables that occasionally hang around and mess up tests.


app/mailers/application_mailer.rb
```diff
@@ -11,4 +12,19 @@ def default_url_options(options = {})
       .merge(locale: I18n.locale)
       .merge(options)
   end
+
+  # Add the image file as an attachment
+  # Typically called by a mailer subclass directly
+  def add_email_signature_logo
+    signature_file = File.read("#{Rails.root}/app/assets/images/#{Otis.config.image.hathitrust.email_signature_logo.name}")
```
mwarin, line 19: this.hurts.to.read.but.thats.rails.i.guess

db/seeds.rb
```diff
@@ -10,6 +10,8 @@
 
 raise StandardError, "Not for production use" if Rails.env.production?
 
+ActiveRecord::Base.connection.execute("DELETE FROM ht_web.otis_approval_requests")
```
mwarin, line 13: This I'm guessing is not directly related to the logo stuff, rather testing, but that's fine.

mwarin (approved):
Looks legit. I see no reason to hold it back. Approve.

I could be convinced to imagine a `Logo` class that hides even more of the internals but the overhead might not be worth it.

app/mailers/application_mailer.rb
moseshll: I was tempted to try `asset_path` but then there's module imports to deal with.

If this kind of thing comes up again we should probably spin off a helper.

PR #209, opened 2024-02-02
Action updates

Ronster2018:


aelkiss (approved):
Looks pretty good to me!

We should squash commits before merging - see some notes on our usual workflow here: https://hathitrust.atlassian.net/wiki/spaces/HAT/pages/2014150657/git+rebase+squash+workflow+for+HathiTrust

aelkiss:
Another thought is that we should probably rename `build-main.yml` to `build.yml`

PR #213, opened 2024-02-29
DEV-1019 Move ht_institutions cron job to OTIS

moseshll:
- Add `InstitutionsExport` class and rake task that invokes it.
- Add three config values:
  - `ht_institutions_file`
  - `ht_saml_entity_ids_file`
  - `export_files_directory`
    - `ht_tanka` is setting `EXPORT_FILES_DIRECTORY` to `/htapps/www/files/tmp/` temporarily, for evaluation.
- Maintenance:
  - Annoying name collision fixed: rename `test/lib/tasks/daily_digest_test.rb` to `daily_digest_task_test.rb`
  - Call `Otis::Application.load_tasks` in test helper to keep rake tests from clobbering each other's coverage

### DIFF
--- a/test/lib/tasks/institutions_export_task_test.rb
+++ b/test/lib/tasks/institutions_export_task_test.rb
@@ -6,7 +6,9 @@ module Otis
   class InstitutionsExportTaskTest < ActiveSupport::TestCase
     def setup
       remove_generated_files
-      create(:ht_institution)
+      5.times do
+        create(:ht_institution)
+      end
     end
 
     def teardown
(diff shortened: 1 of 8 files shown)


test/lib/tasks/institutions_export_task_test.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+module Otis
+  class InstitutionsExportTaskTest < ActiveSupport::TestCase
```
aelkiss, line 6: Probably would be worth checking the content of the files (at least one line of it) -- do we have fixture data in the institutions table?

aelkiss (commented):
All makes sense; wondering if we should test the contents of the generated files though.

aelkiss (approved):
New test looks fine.

PR #227, opened 2024-08-01
DEV-1241 SSD Proxy Reports from PT

moseshll:
This PR adds a new page to Otis: SSD Proxy Reports, based on a facility hidden away in PageTurner. The existing facility uses some of the advanced features of Bootstrap Tables that have been replicated here, and (one hopes) improved upon. Otis uses Bootstrap Tables but until now has not taken advantage of advanced search features, pagination, data export, or a server-side provider (via `format=json`).

Most of the changes in this PR are boilerplate - two new database tables, seeds and factories for them, localizations.

## For the Reviewer
- Proceed with this branch locally as described in the README by building and running the `web` service.
- Please evaluate the basic functionality at `http://0.0.0.0:3000/useradmin/ht_ssd_proxy_reports`, logging in as `admin@default.invalid`. The behavior should at least resemble https://babel.hathitrust.org/cgi/pt/reports/ssdproxy
- The controller at `app/controllers/ht_ssd_proxy_reports_controller.rb` has fairly extensive notes and comments. Do they make sense?
- In future we may want to extend this approach (Bootstrap Table with server-side data source and per-column search controls). Poorly-named data or methods will make this less intuitive so please object to anything that violates the Principle of Least Astonishment.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -92,6 +92,8 @@ gem "flag-icons-rails"
 gem "rails-i18n", "~> 6.0.0"
 gem "maxmind-geoip2"
 gem "jira-ruby"
+gem "ransack"
+gem "kaminari"
 
 gem "canister"
 gem "ettin"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -181,6 +181,18 @@ GEM
       thor (>= 0.14, < 2.0)
     json (2.6.2)
     jwt (2.5.0)
+    kaminari (1.2.2)
+      activesupport (>= 4.1.0)
+      kaminari-actionview (= 1.2.2)
+      kaminari-activerecord (= 1.2.2)
+      kaminari-core (= 1.2.2)
+    kaminari-actionview (1.2.2)
+      actionview
+      kaminari-core (= 1.2.2)
+    kaminari-activerecord (1.2.2)
+      activerecord
+      kaminari-core (= 1.2.2)
+    kaminari-core (1.2.2)
     listen (3.0.8)
       rb-fsevent (~> 0.9, >= 0.9.4)
       rb-inotify (~> 0.9, >= 0.9.7)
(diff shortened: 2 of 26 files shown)


mwarin (approved):
## `https://github.com/hathitrust/otis/pull/227`

builds ok

## Functional test of ssdproxy bootstrap table

Almost all positive feedback below, except when marked "Negative".

The dropdowns, for "rights code" etc, work both when selecting with mouse and typing.

Negative: The input-filter for "Time" does not work as I expect. I type in something, e.g. "01" in the "mm" part, the input field refreshes to say "mm/dd/yyyy" again, the table seems to not do anything.
Once I start typing I only have a half second or so before the input field refreshes, to seemingly no effect.

The text filters (other than "time") do a competent substring filter and reset as expected when set to "".

The text filters are navigable by keyboard as well, which is nice.

Negative: The table contents is much wider than the element it is in, meaning I only see the columns "Time" - "Content Provider" and have to scroll right to see rightmost 5 columns ("Digitization Agent" - "Rights Date").

Pagination works. As with all other table controllers, does not update the URL or remain on the same pagination after refreshing the page, which is fine here but might not do for a public interface.

Columns are sortable and pagination respects the field currently used for sorting.

Changing "rows per page" updates the pagination competently.

Emails and institution names are clickable and take you to the user or institution page.

I want to say the table does everything correctly that I would expect, except the "Time" input.

Check: Exporting data with the "Export data" button works. I got served invalid xml one time but lost track of which search did it.

I can hide/show controls and hide/show columns.

"Clear filters" only clears the "Date range" filter. I expected it to clear all filters.

## `app/controllers/ht_ssd_proxy_reports_controller.rb`

No remarks, looks good to me.

## "boilerplate"

I scanned through the rest of the code changes and nothing jumped out.

table-too-wide-for-its-element

Only minor nitpicks, nothing to stand in the way of an APPROVE.

PR #237, opened 2024-11-18
Bootstrap 5

moseshll:


app/presenters/ht_registration_presenter.rb
```diff
@@ -167,7 +167,7 @@ def show_inst_id
   end
 
   def show_mfa_addendum
-    mfa_addendum ? "<span class='label label-success'><i class='glyphicon glyphicon-lock'></i></span>" : ""
+    mfa_addendum ? "<span class='badge bg-success'><i class='bi bi-lock-fill text-light'></i></span>" : ""
```
carylwyatt, line 170: This slipped my mind when we were working on icons, but if you want them to be accessible, they should have an `aria-label` and probably also `aria-hidden="true"`.

carylwyatt (approved):
I left a comment about aria labels for the bootstrap icons (see [a11y section](https://icons.getbootstrap.com/#accessibility) of icon docs), but I don't think that's a blocker for the PR. Looks good otherwise!

app/presenters/ht_user_presenter.rb
```diff
@@ -82,8 +82,8 @@ def show_expires
       fmt = I18n.l(self[:expires].to_date, format: :long) +
         "<p>#{expiration_badge}</p>"
       if expiring_soon?
-        fmt += "<p><span class=\"bg-danger text-danger\">" +
-          time_to_expiration + "</span></p>"
+        fmt += "<strong><span class=\"text-danger\">" +
+          time_to_expiration + "</spam></strong>"
```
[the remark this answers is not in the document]
moseshll: Need a joke RFC for a `<spam>` tag. Fixed.

app/presenters/ht_registration_presenter.rb
moseshll: @carylwyatt Good. I've got our two uses of the lock icon sorted, I think. What about our use of the trash icon in the "clear date filters" part of the SSD Proxy Reports page, since the icon is part of a button that has a title? Would it still benefit from `aria-hidden`?

```
<button class="btn btn-outline-secondary"
            id="action-clear-datetime-filter"
            onclick="clearDateRangeFilter();"
            title="<%= t ".clear_filter" %>"
          >
          <i class="bi bi-trash"></i>
</button>
```

carylwyatt: Oh interesting! `title` isn't a valid attribute on a button element, so I'm wondering if maybe that's a bootstrap-table thing? [MDN docs suggest](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#icon_buttons) adding the text to button itself but then using CSS to hide that text. Bootstrap has a [`visually-hidden` class](https://getbootstrap.com/docs/5.3/getting-started/accessibility/#visually-hidden-content) that handles that.

carylwyatt: Eh, I take it back, [`title` is considered a global attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title). I'm not sure it's achieving a11y in the button context, though.

moseshll: Okay, realizing I am too much enamored of my "clever" presentation... how 'bout this?
(Changed to "Clear Filters" after screenshot was captured.)

Screenshot 2024-11-19 at 2 57 46 PM

```
<button class="btn btn-outline-secondary"
            id="action-clear-datetime-filter"
            onclick="clearDateRangeFilter();"
          >
          <%= t ".clear_filter" %>
          <i class="bi bi-trash" aria-hidden="true"></i>
</button>
```

carylwyatt: It's beautiful! 🤩

carylwyatt: fwiw, I'm all for "sentence case" in a UI vs. "title case" (Clear filters vs Clear Filters)... but HT doesn't have a clear style guide on that, so it's dealer's choice at the moment.

PR #243, opened 2025-01-07
Remove unnecessary query to try to speed up ht_users index page.

moseshll:
- `@all_users` was being populated via SQL in all cases even though it was only used for CSV download.
- The query has been removed and the CSV now uses the active + inactive arrays so the query was redundant anyway.
- The controller now uses `includes` instead of `joins` to pick up the institution and approval request data for each user.
  - Presenter can just call `ht_approval_request.first` since the controller's query applies the `ORDER BY` we want. This eliminates one additional query per user.


aelkiss (approved):
I don't have a sense of whether this will provide any speedup, but the changes look reasonable to me.

moseshll:
MS indicated the version in staging loaded without a timeout, and the logs for the index page are no longer cluttered with one-off `SELECT`s. It's still not as fast as it probably would be with pagination.

PR #244, opened 2025-01-27
Remove last vestiges of wait-for

moseshll:
- Remove invocations of `wait-for` and the `wait-for` binary
- ~~Remove redundant `mariadb-dev` service and make sure everything is talking to `mariadb-test`~~
  - In the Rails world trying to synchronize when the database gets seeded/purged gets hairy, thus we will keep the dev/test database hierarchy.
  - I would like to know why both databases are needed for running `web` -- prolly something to do with Checkpoint and/or Keycard, not willing to die on that hill today.
- Use `selenium/standalone-chromium` for system tests and remove `docker-compose.yml.arm64`


Ronster2018 (approved):
The changes here have been reviewed and checked out locally. I think we're good to merge this one in.

PR #246, opened 2025-02-28
Rails 8 second go

moseshll:


aelkiss:
@moseshll still wants to exercise the basic code a bit more when all the JS is working.

aelkiss:
@moseshll Should we close as superceded by #247 ?

PR #247, opened 2025-03-06
Rails 8 js

moseshll:
Reviewers:
- Do the revised installation procedure in the README work without additional steps?
- Does the additional step `docker compose up -d web` bring up a healthy web service?
- Do the test procedures `docker compose run --rm test` and `docker compose run --rm system-test` succeed?
- Does the structure of the `Dockerfile` make sense for production?
- Are there problems or potential problems in the simplistic `npm` workflows we have at this point?


carylwyatt:
I didn't make it very far through step 1 😭  
I pulled down the latest changes and:
- no problem with `docker compose build`
- got an error with `docker compose run --rm dev bin/setup-dev.sh`: `line 5: docker: command not found` 🤨 

I guess it doesn't like the docker command in `setup-dev.sh`??

moseshll:
@carylwyatt D'oh, I should have caught that. The command is just ./bin/setup-dev.sh, it contains Docker commands _rather than_ running under Docker. I will update the README.

carylwyatt (approved):
After the README update, everything built and is healthy! 🎉 
Tests ran no problem.
I don't know enough about docker to give you a good answer, but it's all working for me!
And I think the way you're handling scripts and npm is fine. npm is very choose-your-own-adventure, and this all seems to be working (as best as it can for shoe-horning these older js libraries into a modern development situation).

Dockerfile
```diff
@@ -1,24 +1,32 @@
...
+################################################################################
+# PRODUCTION                                                                   #
+################################################################################
 FROM base AS production
 
 ENV RAILS_SERVE_STATIC_FILES true
 ENV RAILS_LOG_TO_STDOUT true
 ENV RAILS_ENV production
+# FIXME: duplicate, see line 7
 ENV BUNDLE_PATH /gems
```
Ronster2018, line 30: My understanding is that this can stay on line 7 and can be removed from line 30. Everything up until line 24 will be used for both containers.

Ronster2018 (changes requested):
The over all structure of the docketfile is fine. I noted the FIXME about the duplicated environment variable for BUNCLE_PATH and this can be left on line 7 and removed from line 30 since everything above the stage for development will be used for both containers.

PR #253, opened 2025-04-04
DEV-1609 manually renewing user causes double renewal

moseshll:
`HTUser#add_or_update_renewal` was designed to update/create/check the audit trail when a user's expiration date is updated. This can happen via outside approval or manually inside Otis (User->Edit->change date). However, it renewed the user (bumped the expiration date) as a side effect. This behavior is correct when an approval request for a supervisor outside HathiTrust is processed. It is not correct when a manual update is done, because the expiration date gets updated twice: once from the HTML form and a second time by this method.

This patch:
- Adds comments about what `HTUser#add_or_update_renewal` does.
- Moves responsibility for calling `HTUser#renew!` to the caller, specifically `HTApprovalRequestsController`.
- Adds one initially failing test that validates expiration date when a user is edited.

### DIFF
--- a/app/controllers/ht_approval_requests_controller.rb
+++ b/app/controllers/ht_approval_requests_controller.rb
@@ -115,7 +115,9 @@ class HTApprovalRequestsController < ApplicationController
     end
     adds = []
     emails.each do |e|
-      HTUser.find(e).add_or_update_renewal(approver: current_user.id)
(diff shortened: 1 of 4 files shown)


mwarin (approved):
The changes match the description, the changes make sense and the tests run green.

Approve!

PR #254, opened 2025-04-11
TTO-301 Otis: investigate possible missing e-mail registration text

moseshll:
In the migration from `ckeditor` 4 to 5 the `textarea` that contains the email body (for approval request and registration emails) lost its `name` attribute, so we sent out some emails that omitted most of the intended content.

- Restore `email_body` form element name to approval request and registration email previews.
  - The changes in this PR that actually fix the bug are in `preview.html.erb` and `edit.html.erb`, everything else is just tests & tidy.
- Mailers now raise if the `email_body` is omitted. Better a 500 than a mostly blank e-mail being sent. This required some modifications to the tests, to make sure we have some fake body text.
- TIDY phase: at the Controller-Mailer interface the `email_body` param was getting renamed to just `body` as a Mailer instance variable. For naming continuity -- to make this bucket chain more understandable -- this is now `email_body` all the way thru. Unfortunately this sneezed a lot of changes into the `diff` for this branch but they can be ignored.

### DIFF
--- a/app/controllers/ht_approval_requests_controller.rb
+++ b/app/controllers/ht_approval_requests_controller.rb
@@ -40,7 +40,7 @@ class HTApprovalRequestsController < ApplicationController
   def update
     begin
       ApprovalRequestMailer.with(reqs: @reqs, base_url: request.base_url,
-        body: params[:email_body], subject: params[:subject]).approval_request_email.deliver_now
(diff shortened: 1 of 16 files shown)


mwarin (approved):
It builds, tests are green and the changes all make sense. 
Approve!

PR #257, opened 2025-07-30
Out of band dependency update 2025-07-30

moseshll:
- Critical nokogiri patch
  - Low severity thor patch

Reviewer: doing this out of band since we're not sure about the release schedule for Resource Sharing changes; I don't want to depend on rolling these updates into ETT-291 or its brethren.


Ronster2018 (approved):
These security updates look good to me.

PR #258, opened 2025-08-21
ETT-413 Remove apparently-unused template field from ht_institutions

moseshll:
- Remove from schema
- Remove from model
- Remove from prosenter
- Remove localizations


aelkiss (approved):
This all looks fine; we should test in staging to make sure we know what order to do things in -- I think it should be OK to deploy before we update the actual schema in production to remove the template column -- but we should verify.

PR #261, opened 2025-09-03
ETT-279 SSD Proxy (ATRS) report in OTIS doesn't have usage data beyon…

moseshll:
…d Oct 15, 2024

- Begin translation of `babel/pt` cron to extract logfile data in Otis
- Begin transition from MiniTest to RSpec


app/lib/otis/log_importer.rb
```diff
@@ -0,0 +1,74 @@
...
+require "zinzout"
+
+module Otis
+  class LogImporter
+    attr_reader :user_map
+
+    def self.logs
+      glob = File.join(
+        ENV.fetch("ULIB_LOGS_ARCHIVE", "/ulib-logs/archive"),
+        "{macc,ictc}*/var/log/babel/access-imgsrv_downloads.log*"
```
aelkiss, line 14: We'll have to rethink this part since we won't be able to mount it directly

app/lib/otis/log_importer.rb
aelkiss, line 1: The old perl version keeps track of the last successful run; we'll want to do something like that or adopt a similar thing as we're doing in metadata processing with the journal.

We'll want to keep some level of idempotence here -- the perl version uses this checksum computed in the database; we could keep the same thing, or we could think about some other way to avoid duplication in the database.

app/lib/otis/log_importer.rb
aelkiss: Then we could do rclone copy or maybe just rclone cat for each relevant log file.

aelkiss:
For integration testing with rclone we can use "versitygw" to serve as a filesystem to s3 gateway and set that up in an rclone config. Some examples for ingest here: https://github.com/hathitrust/feed/tree/ETT-574.

Config using that example from docker-compose:

access key TESTINGACCESSKEY
secret key testingsecretkey
endpoint http://versitygw:7070

keep everything else the default, make some directories under `var/vgw` (that's the default path it's mounting) and it should mostly work modulo any filesystem permission weirdness (just make sure versitygw is running as the same user as the container any integration tests trying to create files under there is, or commit the files ahead of time and make sure versitygw at least has read access)

spec/lib/otis/log_importer_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+    it "ignores entries earlier than the last run file" do
+      # Write a last run file with the current date
+      File.open(importer.last_log_import_path, "w") do |file|
+        file.puts Time.now
+      end
+      # ... and travel into the future
+      Timecop.travel(Time.now + 7.days) do
+        expect do
+          importer.run
+        end.to change { HTSSDProxyReport.count }.by(0)
```
moseshll, line 37: FIXME: This is not a good test. Should keep stats on the number of files read and/or the number of entries inserted, since the uniqueness constraints AND the date math can both keep the count from changing. The test should observe the effects most closely related to the time travel.

app/lib/otis/log_importer.rb
moseshll: I'm keeping it the same 20-byte binary for ETT-279 but I have a note in the model file that a change in the schema would simplify things. The calculation is being done in Ruby now.

app/lib/otis/log_importer.rb
```diff
@@ -0,0 +1,163 @@
...
+          rescue JSON::ParserError
+            Rails.logger.error "unparseable JSON in #{log_file} line #{line_number + 1}"
+            next
+          end
+          if relevant_log_entry?(entry)
+            # Call `.save` and not `.save!` on this object because its SHA may
+            # violate uniqueness, which is perfectly okay.
+            report = create_report(entry)
+            @stats[:entries_found] += 1
+            if report.save
```
aelkiss, line 103: Would it clean this up if we swallowed the exception if there's a duplicate key error based on the sha, and log it otherwise? Failing that, maybe we should extract this logic to a separate method?

aelkiss (approved):
All makes sense to me with a quick look-through; basically doing what I was expecting, and the tests look good. I did note one area I wonder if we could clean up and one question just to verify if we made sure it will work with the SMB/CIFS back end.

I do see a couple of places there will be conflicts with https://github.com/hathitrust/otis/pull/264 specifically with changes to the table & migrating its tests to rspec, but that shouldn't be a big deal to fix.

app/lib/otis/log_transfer.rb
```diff
@@ -0,0 +1,54 @@
...
+    def imgsrv_logs
+      @imgsrv_logs ||= begin
+        cmd = <<~RCLONE.gsub(/\s+/, " ").strip
+          rclone
+          --config #{rclone_config_path}
+          lsjson
+          -R
+          --files-only
+          --no-mimetype
+          --include '/{macc,ictc}-ht-web-*.umdl.umich.edu/var/log/babel/access-imgsrv_downloads.log*'
```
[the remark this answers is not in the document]
moseshll: Yes, the command was iteratively developed inside k8s. The --include flag was the tricky one and needed the real world experimentation.

app/lib/otis/log_importer.rb
moseshll: Unfortunately if any validation fails (from `save!`) it seems to raise `ActiveRecord::RecordInvalid` and I have to look at the errors anyway. I am moving this down out of the main loop however.

app/models/ht_ssd_proxy_report.rb
```diff
@@ -30,4 +31,24 @@ def hf
...
+
+  def sha
+    self[:sha] = calculate_sha
+  end
+
+  private
+
+  # Should be equivalent enough to what was previous being done inside MariaDB:
+  # UNHEX(SHA1(CONCAT_WS(' ', `datetime`, `htid`, `in_copyright`, `is_partial`, `email`, `inst_code`)))
+  # TODO: ETT-298 may alter the table schema at which time this can be revisited. It may be
```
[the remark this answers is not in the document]
moseshll: I don't think anyone will suffer if we leave this as binary 20 with an unhex. I am going to get rid of the explicit TODO.

PR #264, opened 2025-10-03
ETT-298: Resource sharing user stats

aelkiss:
This implements most of what we need to *display* resource sharing user stats.

* renames “SSDProxyReport” as “Download”
* renames the table for tracking to otis_downloads
* adds role, partial?, page_count on the model & tests for this
* ~routes `/download/:role` for ssdproxy and resources sharing users separately~
* adds display for role, page counts

~This does *not* incorporate changes to data loading (#261) to account for this -- hence why it's in draft. Once we merge #261 we can rebase this branch against that and add additional changes to data loading to account for the new columns & table name.~

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -258,7 +258,7 @@ GEM
     puma (7.0.4)
       nio4r (~> 2.0)
     racc (1.8.1)
-    rack (3.2.1)
+    rack (3.2.2)
     rack-session (2.1.1)
       base64 (>= 0.1.0)
       rack (>= 3.0.0)


app/helpers/application_helper.rb
```diff
@@ -2,9 +2,17 @@
...
+    %w[users approval_requests institutions contacts contact_types logs registrations]
       .select { |item| can?(:index, "ht_#{item}") }
       .map { |item| {item: item, path: send(:"ht_#{item}_path")} }
+      .concat(download_nav_items)
+  end
+
+  def download_nav_items
+    return [] unless can?(:index, :ht_downloads)
+
+    %w[ssdproxy resource_sharing]
```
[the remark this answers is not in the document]
aelkiss: The more I think the more I really want it on the same page & filterable. I'll push a commit with that, but we can back it out if we decide that separate pages is preferable.

app/presenters/ht_download_presenter.rb
```diff
@@ -93,4 +97,15 @@ def show_institution_name
 
     link_to institution_name, ht_institution_path(inst_code)
   end
+
+  def show_pages
```
aelkiss, line 101: See comments elsewhere about pages & Ransack; not sure what the best option here is.

config/locales/en.yml
```diff
@@ -340,12 +342,16 @@ en:
...
+  ht_download:
     index:
       clear_filter: Clear Filters
       date_range: Date Range
       filter: Filter
-      ht_ssd_proxy_reports: SSD Proxy Reports
+      download_reports: Download Reports
+    values:
+      role:
+        resource_sharing: Resource Sharing
```
aelkiss, line 353: I tried adding this to localize the values showing up on the reports page, but there are two issues:
1) it looks terrible because it wraps
2) filtering by that column no longer works, again because the values getting passed for filtering have been localized

@moseshll any thoughts? Would it be feasible to add something to pass the original (non-presented) value back through for filtering, or is there a better way to do it?

aelkiss:
Also wondering @moseshll if I should be migrating any tests I touch to rspec after we merge #261 ?

app/presenters/ht_download_presenter.rb
moseshll: To the `ht_download` model you can start with a custom ransacker to translate the special case value of NULL in the table.

```
ransacker :pages
  Arel.sql("COALESCE(#{table_name}.pages, 'all')")
end
```

It doesn't take into account the possible "unknown" case represented in the presenter, where it's a partial download. That would be a more complex Arel operation (using Arel's equivalent of SQL `IF` perhaps, in place of the `'all'` literal?).

app/controllers/ht_downloads_controller.rb
```diff
@@ -38,11 +39,13 @@ class HTSSDProxyReportsController < ApplicationController
     "institution_name" => :ht_institution_name_i_cont,
     "rights_code" => :ht_hathifile_rights_code_eq,
     "rights_date_used" => :ht_hathifile_rights_date_used_eq,
-    "title" => :ht_hathifile_title_i_cont
+    "title" => :ht_hathifile_title_i_cont,
+    "pages" => :pages_eq
```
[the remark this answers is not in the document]
moseshll: See below on the custom ransacker. I would ask AZ for guidance on one vs two columns. I think we can support either approach.

config/locales/en.yml
moseshll: Fix the wrapping with a presenter (like we do with the date)
  ```
    def show_role
      "<span class=\"text-nowrap\">#{localize_value(:role)}</span>"
    end
  ```
  This will just get you the English localization because in index.html.erb there's no locale passed back to the JSON endpoint. Can stop there, I think.

OTOH, if you want localized results, change `data-url` in the Bootstrap Table setup:
  ```
  data-url="ht_downloads?format=json&locale=<%= I18n.locale %>"
  ```

And then there's additional stuff that can be done to get the presenter to tell Bootstrap Table how to map between search and display values for the column. I am halfway thru mocking it up but since we are going to step back on localization I think this would be a bridge too far.

To shut up `i18n-tasks health` (we'll give it the heave-ho maybe later when we address the future of Otis i18n), add to the ignore list in `config/i18n-tasks.yml` around line 115
```
  - 'ht_download.values.*'
```

It complains because it doesn't see any of these in the ERB templates. Duh, we're using them dynamically. i18n-tasks is a bit of a dimwit out of the box.

aelkiss:
Great. I think I'm inclined to at least get everything passing (if not working 100% as expected) in this branch, then rebase against https://github.com/hathitrust/otis/pull/261 (especially if we can get that merged soon), then resolve the other issues.

PR #288, opened 2026-05-18
ETT-1356 Otis: move Jira credentials from Rails to sealed secret

moseshll:
- Adds to `Otis::JiraClient` a `credentials` class method to fetch credentials from a YAML file.
  - `OTIS_JIRA_CONFIG` will provide the path in k8s, in Docker we provide `config/jira.yml` with bogus defaults.
- Remove Jira credentials from Rails encrypted file.

I'm not fussed about the Coveralls failures. We are transitioning from Minitest to Rspec so there are going to be some minor fluctuations in coverage.

We also plan (see ETT-533) to do a proper webmock test setup, so the janky NullClient will go away in due time.


app/lib/otis/jira_client.rb
```diff
@@ -26,6 +29,13 @@ def self.create_client
       end
     end
 
+    # Read YML file at OTIS_JIRA_CONFIG or config/jira.yml
+    # and return a hash with :username and :password
+    def self.credentials
```
moseshll, line 35: Not interested in memoizing this even if we have it called twice in rapid succession. Sensitive data here, best to let it be GCed.

eumalin (approved):

eumalin (commented):
Might be worth adding more test coverage, e.g. for the OTIS_JIRA_CONFIG variable path

eumalin (approved):

PR #290, opened 2026-07-06
July dependency updated

liseli:
```
bundle update --bundler
bundle update --all
npm update
npm audit fix --force --> It was applied in the last update to get the breaking CKEditor changes due to the XSS vulnerability
```

Apply `bundle exec standardrb --fix` after GitHub tests failed.


aelkiss (changes requested):
We need to change this in `Gemfile`:

```
source: "https://rubygems.org"
```

```
source: "https://rubygems.org", cooldown: 7
```

Can you please make that change, revert the `Gemfile.lock` to its previous state, and then do `bundle update --all` again?

Thanks!

aelkiss (approved):
Ruby part looks good to me. I'd still like @moseshll to take a look at the javascript part.

moseshll (approved):
Ckeditor seems to working fine. That said, I will be happy when we can give it the heave-ho and have Jira do the heavy lifting for us.

PR #292, opened 2026-07-08
ETT-1616 Show only service users are registered for

moseshll:
ETT-1616 Show only service users are registered for, not user type, role, and access in interface

 - The `ServiceRole` class has been expanded to map `HTRegistration.role` roles to legacy-ish `HTUser.role`, and with as much description as possible.
 - Automatic assignment of `ht_users.access` and `ht_users.usertype` based on role.
 - User type and access can now be removed from the UI.


test/presenters/ht_user_presenter_test.rb
```diff
@@ -4,10 +4,8 @@
 
 class HTUserPresenterTest < ActiveSupport::TestCase
   test "class constants" do
-    assert_not_nil HTUserPresenter::ALL_FIELDS
-    assert_equal 15, HTUserPresenter::ALL_FIELDS.count
-    assert_not_nil HTUserPresenter::INDEX_FIELDS
-    assert_equal 8, HTUserPresenter::INDEX_FIELDS.count
+    assert HTUserPresenter::ALL_FIELDS.count.positive?
```
moseshll, line 7: These tests were too rigid.

test/presenters/ht_user_presenter_test.rb
```diff
@@ -108,22 +106,10 @@ class HTUserPresenterTest < ActiveSupport::TestCase
 class HTUserPresenterRoleDisplay < ActiveSupport::TestCase
   def setup
     @user = HTUserPresenter.new(build(:ht_user, role: "crms"))
-    @user_unknown_role = HTUserPresenter.new(build(:ht_user, role: "blah"))
```
moseshll, line 111: `HTUser` no longer allows underspecified role so these became obsolete.

test/lib/otis/service_role_test.rb
```diff
@@ -1,19 +0,0 @@
-# frozen_string_literal: true
```
moseshll, line 1: Migrated to rspec.

test/controllers/ht_users_controller_test.rb
```diff
@@ -259,7 +258,7 @@ def setup
       "userid,displayname,email,activitycontact,approver," \
       "authorizer,usertype,role,access,expires,expire_type," \
       "iprestrict,mfa,identity_provider,inst_id,inst_name", @response.body
-    assert_match "a@b,A B,c@d,e@f,g@h,i@j,staff,ssd,total,2020-01-01 00:00:00 UTC," \
+    assert_match "a@b,A B,c@d,e@f,g@h,i@j,student,ssd,normal,2020-01-01 00:00:00 UTC," \
```
moseshll, line 261: `staff/ssd/total` is just wrong. The `ssd` role above is now sufficient to generate the correct user type and access.

spec/models/ht_download_spec.rb
```diff
@@ -5,9 +5,7 @@
 
   around(:each) do |example|
     described_class.delete_all
-    ClimateControl.modify(TEST_TMP: "") do
```
moseshll, line 8: Copypasta from holdings? Not needed -- we don't use `TEST_TMP` anywhere in Otis.

spec/lib/otis/jira_client/registration_spec.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+RSpec.describe Otis::JiraClient::Registration do
```
moseshll, line 3: Our Jira automations depend on some of these parameters being specific values, so I wrote tests for these bits as a defense against UI modifications (display values) messing with the expectations we have for Jira inputs.

spec/lib/otis/registration_mover_spec.rb
```diff
@@ -2,32 +2,20 @@
 
 RSpec.describe Otis::RegistrationMover do
   describe "#ht_user" do
-    [:ssdproxy, :resource_sharing].each do |role|
-      context "with #{role} role" do
-        it "has 'normal' access" do
+    Otis::ServiceRole.keys.each do |role_key|
+      context "with service role #{role_key}" do
+        it "creates a user with an expected user_type, access, and role" do
```
moseshll, line 11: Well, it doesn't really check correctness, just "expected" in that they're attested values. Bumping up against `expect(1).to eq(1)` territory here.

config/locales/en.yml
```diff
@@ -38,7 +38,7 @@ en:
         pages: Pages
         rights_code: Rights Code
         rights_date_used: Rights Date
-        role: Role
+        role: Service Role
```
moseshll, line 41: We're gradually abandoning localization, so the changes under `locales` are mostly deletions.

app/presenters/ht_registration_presenter.rb
```diff
@@ -241,6 +241,6 @@ def expire_type_options
   end
 
   def role_options
-    @role_options ||= HTRegistration::ROLES.sort.map { |role| [I18n.t("ht_registration.values.role.#{role}"), role] }
+    @role_options ||= HTRegistration::ROLES.sort.map { |role| [Otis::ServiceRole.new(role).full_name, role] }
```
moseshll, line 248: In a sense, this line is the heart of the PR.

app/models/ht_user.rb
```diff
@@ -51,6 +57,17 @@ def resource_id
     email
   end
 
+  def role=(new_role)
```
moseshll, line 60: Instead of assigning access and usertype independently, now `role` uniquely determines them. The `@service_role = nil` is to break memoization.

app/models/ht_user.rb
```diff
@@ -13,9 +13,10 @@ def initialize(msg = "Renewal Error", type:)
 end
 
 class HTUser < ApplicationRecord
-  ROLES = %i[corrections cataloging ssdproxy crms quality staffdeveloper staffsysadmin replacement ssd resource_sharing].freeze
```
moseshll, line 16: This enumeration is _ancient_, with obsolete roles like `staffsysadmin`. Great that we can have an authoritative list, but without `ServiceRole` as the source of truth this is semantically bleached.

app/models/ht_registration.rb
```diff
@@ -11,9 +11,7 @@ class HTRegistration < ApplicationRecord
   alias_attribute :approved, :finished
   alias_attribute :submitted, :received
 
-  # FIXME: the roles in HTUser are expected to be simplified to look more like this.
```
moseshll, line 14: A prescient FIXME

app/lib/otis/registration_mover.rb
```diff
@@ -11,12 +11,17 @@ def ht_user
 
       institution = HTInstitution.find(@registration.inst_id)
       @ht_user = @registration.existing_user || HTUser.new(email: @registration.applicant_email)
-      @ht_user.update(userid: userid, displayname: @registration.applicant_name,
```
moseshll, line 14: The reformatting masks the fact that we're now omitting `access` and `usertype` and letting `HTUser` + `ServiceRole` figure that part out.

app/lib/otis/jira_client/registration.rb
```diff
@@ -111,6 +111,18 @@ def ea_fields
       end
     end
 
+    def ea_field_summary
```
moseshll, line 114: To make testing easier, see subsequent comment on Jira needing certain values to run automations correctly.

app/lib/otis/service_role.rb
```diff
@@ -1,56 +1,96 @@
...
+  # when a user is created.
+  # Maybe we can just go with the new SERVICE_ROLES.keys values in ht_users.role sooner or later
+  # but care must be taken since the babel Perl code relies on certain legacy values.
+  # TODO: consider moving this data to YML?
 
-    ROLE_TO_SERVICE_NAME = {
-      crms: CAA_NAMES,
-      quality: CAA_NAMES,
+  class ServiceRole
+    SERVICE_ROLES = {
```
moseshll, line 15: As much semantics as I could stuff into the various roles. We're not doing localization, or more accurately we're undoing localization. Sick of the additional layer of indirection that entails. So having UI visible strings here is fine.

app/lib/otis/registration_mover.rb
```diff
@@ -29,14 +34,9 @@ def ht_user
...
-    # ssdproxy role grants normal access, all other roles grant total access
-    def access
-      case @registration.role
-      when "resource_sharing", "ssdproxy"
-        "normal"
-      else
-        "total"
-      end
+    # Map `ServiceRole` key to legacy `ht_users.role`
+    def user_role(service_role)
```
eumalin, line 39: Heads up, this throws now instead of just returning something harmless. Nothing catches it, not even in approve. Any old-vocab role sitting in the DB (quality, staffdeveloper, whatever) and staff clicking approve gets a 500. Do we need to consider backfilling those rows, or should this handle unknowns itself?

app/models/ht_registration.rb
eumalin: This is still checking ssdproxy but ATRS is atrs now. So this exemption just never fires for ATRS anymore, they're stuck requiring hathitrust_authorizer for no reason. Do we need to look at this?

app/lib/otis/registration_mover.rb
eumalin: Same ssdproxy typo by omission as the other file, and it flips this condition too, ATRS users get hathitrust_authorizer instead of always falling back to auth_rep_email. Feels like both of these should just ask ServiceRole instead?

app/lib/otis/registration_mover.rb
moseshll: I *think* we're going to make sure there are no old-style roles in the DB before deploying this. That part of the requirements seems to be a bit in flux but I'd say it most likely the registrations will be updated. I'm going to leave this unresolved until I know for sure.

app/lib/otis/registration_mover.rb
moseshll: I've asked Ange if there is a common moniker we can use for `ssd` + `ssdproxy` and give some semantics to the business logic and find a place for it in `ServiceRole`. At the very least I have a patch for this.

app/models/ht_registration.rb
moseshll: Should be fixed.

eumalin (approved):
LGTM

aelkiss:
I have rebased again. Both branches had lots of changes in `db/seeds.rb` with a different implementation for creating expired users, but I believe I've kept the essence of the change that was here:
* creating 50 expired users and 100 non-expired users
* using legacy roles for expired users

aelkiss:
@eumalin I've rebased this branch; you had reviewed back in July but it's probably worth another look now.

app/lib/otis/registration_mover.rb
eumalin: Did we confirm there are no old style roles left in the DB? If that's not confirmed yet, what do you think about  either confirming it before merging, or adding a rescue for UnknownRoleError so an old role doesn't 500 the approve action?

spec/lib/otis/registration_mover_spec.rb
eumalin: ht_registration.rb was fixed, but I don't see the same fix here yet. Did the patch for this land, or is it still pending?

aelkiss: There are no pending registrations with the old names, but I am a bit confused about what the plan is here and why we have different role keys for the registrations compared to the users table, and I think @moseshll needs to weigh in on that.

moseshll: I've pushed a commit that aligns the role values used in `otis_registrations` with the values in `ht_users`. The rationale (if you can call it that) for the previous arrangement is commented on, in ETT-1887. This should mostly eliminate the "legacy role" presentation seen in preview. Will build and deploy to preview.

moseshll: Deployed to otis-preview.

app/models/ht_registration.rb
```diff
@@ -11,7 +11,9 @@ class HTRegistration < ApplicationRecord
   alias_attribute :approved, :finished
   alias_attribute :submitted, :received
 
-  ROLES = Otis::ServiceRole.keys.freeze
+  # Registrations use the same kind of "semi-legacy" role values as `HTUser`.
+  # i.e., role names that may not match the service role names e.g. "ssdproxy" == "ATRS"
+  ROLES = Otis::ServiceRole.keys.map { |role_key| Otis::ServiceRole.new(role_key).role.to_s }.freeze
```
aelkiss: There are a few things of this form both here and in `HtUser`; we could move these to be in `ServiceRole`

aelkiss (approved):
All makes sense to me. I appreciate this doesn't change the keys `Otis::ServiceRole`, just the roles in `HtRegistration`. We could consider a little bit of cleanup for these things in `HtRegistration` and `HtUser` that are manipulating the data in `Otis::ServiceRole`, but not critical.

app/models/ht_registration.rb
moseshll: Did this and cleaned up some stuff in the presenters too.

PR #299, opened 2026-07-28
Out of band dependency updates July 2026

moseshll:
Out of band dependency updates for a couple of high severity issues in Ruby and JS dependencies.


aelkiss (approved):
All looks fine to me, mostly tiny version bumps; should check in staging before deploying to production.

